PR #3, opened 2019-03-18
loader: Don't cross 32/64 registry silos

daveh-lunarg:
Remove search of both 32- and 64-bit keys when locating runtime on Windows.


daveh-lunarg:
aka https://gitlab.khronos.org/openxr/openxr/merge_requests/1180

PR #9, opened 2019-03-19
Make vulkan SDK dependency to be optional

yl-msft:


daveh-lunarg (approved):
This looks like a good change - not sure why Vulkan was an unconditional requirement.

I added the same conditional Vulkan lib for helloxr on Linux, and built it successfully on a Vulkan-less box.

daveh-lunarg:
Added a commit to do the same conditional Vulkan link on linux systems.  HelloXR builds fine on a linux system without Vulkan, and I think it'll run on a system with just OpenGL - but not 100% certain.  If anyone knows that Vulkan linkage _should_ be required unconditionally on linux, chime in and I'll remove my commit from this PR.

I'll let this simmer a couple days to see if anyone weighs in, then push it. @yl-msft, I assume there's no time pressure to merge this right away?

rpavlik:
Fwiw, I don't know about building, but it definitely runs fine in a vulkan only mode (no opengl) on Linux, that was most of our demos 😁

yl-msft:
@daveh-lunarg  should this change be merged?

PR #13, opened 2019-03-22
Do not query GL context API version before creating context

pH5:
Depending on the platform, querying GL API version without a valid context may do nothing. Move the API version check after context creation and initialize major and minor variables to avoid a comparison with uninitialized values.


daveh-lunarg (approved):
This looks right to me.

PR #14, opened 2019-03-24
Initialize XrExtensionProperties array parameter for rt_xrEnumerateIn…

pH5:
…stanceExtensionProperties

Initialize the `XrExtensionProperties` `type` and `next` fields before passing them to the runtime `xrEnumerateInstanceExtensionProperties` implementation.


daveh-lunarg (approved):
Right, initialization is required.  Good catch!

PR #16, opened 2019-03-25
Add install target to cmake files

jherico:
Addresses #16 

This is a minimal implementation that uses the CMake `INSTALL` directive to install the built binary, generated headers, license file and the generated loader helper files relative to an installation directory.  

Tested with `cmake .. -DCMAKE_INSTALL_PREFIX=F:/openxr_install_test`  

Running the INSTALL target in visual studio resulted in the following files being created in the output folder.

```
./include/openxr
./include/openxr/openxr.h
./include/openxr/openxr_platform.h
./include/openxr/openxr_platform_defines.h
./lib/openxr_loader-0_90.lib
./LICENSE
./src/xr_generated_loader.cpp
./src/xr_generated_loader.hpp
```

Doing this will allow me to create a PR to integrate OpenXR into vcpkg much more easily.  Any suggestions for additional files that should be treated as installable artifacts or better idiomatic cmake usage would be gratefully welcomed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,3 +40,5 @@ if(BASH_COMMAND)
             USES_TERMINAL)
     endif()
 endif()
+
+install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE" DESTINATION share/doc/openxr)
--- a/include/openxr/CMakeLists.txt
+++ b/include/openxr/CMakeLists.txt
@@ -63,3 +63,12 @@ set_source_files_properties(
     ${CMAKE_CURRENT_BINARY_DIR}/openxr_platform.h
     PROPERTIES GENERATED TRUE
 )
+
+set(INSTALL_HEADERS 
+    ${CMAKE_CURRENT_BINARY_DIR}/openxr_platform_defines.h
+    ${CMAKE_CURRENT_BINARY_DIR}/openxr.h
+    ${CMAKE_CURRENT_BINARY_DIR}/openxr_platform.h)
+
+INSTALL(FILES ${INSTALL_HEADERS} 
+    DESTINATION include/openxr
+)
--- a/src/loader/CMakeLists.txt
+++ b/src/loader/CMakeLists.txt
@@ -162,6 +162,8 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
     endif()
 endif()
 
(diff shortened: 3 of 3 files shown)


Ralith:
`LICENSE` loose in the install prefix is unconventional, as is the `src` subdir. Normally neither of those types of things would be installed at all. Should the .cpp be compiled into a library instead, and the header placed in include?

jherico:
As stated, the point of the PR is to make OpenXR usable via package managers.  In my case I manage most of my project's external dependencies with vcpkg.

> LICENSE loose in the install prefix is unconventional

I'm perfectly happy to put it or name it something else (perhaps `share/docs/openxr/LICENSE.txt`?) .  However, one of the requirements for vcpkg is the existence of a license or copyright file that can be installed.  

As for the cpp library, it's described as an optional helper library.  Compiling it into the existing library target is a non-starter since we don't want to introduce any C++ specific dependencies into a library that might need to run in a C-only environment.  Providing it in source form seemed like a reasonable compromise in order to make it usable to people who might be using openxr via a package manager.

Ralith:
> As stated, the point of the PR is to make OpenXR usable via package managers.

That is my concern as well. I maintain the Vulkan packages on NixOS, and plan to maintain OpenXR packages once they stabilize.

> I'm perfectly happy to put it or name it something else (perhaps share/docs/openxr/LICENSE.txt?) . 

I think `share/doc/openxr/LICENSE` (note the singular `doc`) is idiomatic.

> As for the cpp library, it's described as an optional helper library. Compiling it into the existing library target is a non-starter since we don't want to introduce any C++ specific dependencies into a library that might need to run in a C-only environment.

So make it a separate, optional, library?

jherico:
@Ralith I've removed the installation of the C++ helper for now, since it's not on my critical path.  I've also changed the license path to `share/doc/openxr/LICENSE` as you suggested.

rpavlik:
Hmm - not sure what c++ helper lib you're referring to - there are some generated utility functions in files you mention in your original PR comment, but those are mostly helpers for runtimes, not apps.

This implementation of the loader is C++, so using it in a c-only environment in any way is a non-starter. Presumably if a C-only environment arises, someone will write a C version.

In any case, thanks for the contribution, and sorry it took so long to merge it!

(fwiw, @jherico , a while ago I hacked together a quick vcpkg file for this using a different permit-install patch, that you're welcome to start from https://github.com/rpavlik/vcpkg/tree/openxr )

PR #17, opened 2019-03-26
Timespec fixes

rpavlik:
These changes are required to allow the loader to supply the timespec-related extension functions to an application. It also fixes the bad code generation associated with the api dump layer when trying to deal with a struct timespec.


daveh-lunarg (approved):
Looks good to me.

PR #18, opened 2019-03-26
Validation fixes

rpavlik:
Gets the validation layer further in actually working with Hello_xr as well as a local headless sample app.

### DIFF
--- a/src/api_layers/core_validation.cpp
+++ b/src/api_layers/core_validation.cpp
@@ -635,19 +635,34 @@ XrResult CoreValidationXrCreateSession(XrInstance instance, const XrSessionCreat
             }
             cur_ptr = reinterpret_cast<const XrBaseInStructure *>(cur_ptr->next);
         }
-        if (num_graphics_bindings_found != 1) {
+        auto const &enabled_extensions = gen_instance_info->enabled_extensions;
+        bool has_headless = (enabled_extensions.end() !=
+                             std::find(enabled_extensions.begin(), enabled_extensions.end(), XR_KHR_HEADLESS_EXTENSION_NAME));
+        bool got_right_graphics_binding_count = (num_graphics_bindings_found == 1);
+        if (!got_right_graphics_binding_count && has_headless) {
+            // This permits 0 as well.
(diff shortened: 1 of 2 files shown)


src/api_layers/core_validation.cpp
```diff
@@ -635,19 +633,34 @@ XrResult CoreValidationXrCreateSession(XrInstance instance, const XrSessionCreat
             }
             cur_ptr = reinterpret_cast<const XrBaseInStructure *>(cur_ptr->next);
         }
-        if (num_graphics_bindings_found != 1) {
+        auto const &enabled_extensions = gen_instance_info->enabled_extensions;
+        bool has_headless = enabled_extensions.end() !=
+                            std::find(enabled_extensions.begin(), enabled_extensions.end(), XR_KHR_HEADLESS_EXTENSION_NAME);
+        bool got_right_graphics_binding_count = num_graphics_bindings_found == 1;
```
daveh-lunarg: Wouldn't mind a set of parentheses to show the precedence explicitly here

src/api_layers/core_validation.cpp
```diff
@@ -635,19 +633,34 @@ XrResult CoreValidationXrCreateSession(XrInstance instance, const XrSessionCreat
...
+        if (!got_right_graphics_binding_count && has_headless) {
+            // This permits 0 as well.
+            got_right_graphics_binding_count = num_graphics_bindings_found == 0;
```
daveh-lunarg: Parens

daveh-lunarg (approved):
Looks good.

Comments for a slight preference to explicitly show precedence on multiple operator statements.

rpavlik:
Sure, those look good to me, will do.

src/api_layers/core_validation.cpp
rpavlik: done! (Also added on the previous line - the lack of a nice `std::contains` function causing that iterator-filled monstrosity has bugged me for a long time...)

rpavlik:
OK, patches revised accordingly.

PR #19, opened 2019-03-27
fix CMAKE_PRESENTATION_BACKEND default on linux

ChristophHaag:
On Linux presentation.cmake defaults the variable PRESENTATION_BACKEND to xlib.
This include needs to happen before the PRESENTATION_BACKEND variable is used.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -33,6 +33,12 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     add_definitions(-DXR_OS_LINUX)
 endif()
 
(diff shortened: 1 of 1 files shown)


rpavlik (approved):
Looks good to me. Not sure if I'm allowed to merge here, though.

daveh-lunarg (approved):
LGTM.  I also don't know if I'm allowed to merge here.  But I've been doing it :)

@rpavlik, I have no reservations about you pushing PRs here, if that's a concern. My thinking is if the 'merge' button is shown to you, you're allowed to push it (more or less).

PR #20, opened 2019-03-27
Fix Linux manifest file search

Ralith:
This fixes various unintended behavior (such as trying to parse json files from completely unrelated applications) and gets the loader working with an out-of-the-box Monado installation.

### DIFF
--- a/src/loader/manifest_file.cpp
+++ b/src/loader/manifest_file.cpp
@@ -637,40 +637,39 @@ XrResult RuntimeManifestFile::FindManifestFiles(ManifestFileType type,
             LoaderLogger::LogErrorMessage("", "RuntimeManifestFile::FindManifestFiles - unknown manifest file requested");
             throw std::runtime_error("invalid manifest type");
         }
-        bool override_active = false;
-        std::vector<std::string> filenames;
-        ReadDataFilesInSearchPaths(type, OPENXR_RUNTIME_JSON_ENV_VAR, "", override_active, filenames);
-        if (!override_active) {
+        std::string filename;
+        const char *override_path = PlatformUtilsGetSecureEnv(OPENXR_RUNTIME_JSON_ENV_VAR);
+        if (override_path != nullptr && *override_path != '\0') {
+            filename = override_path;
+            std::string info_message = "RuntimeManifestFile::FindManifestFiles - using environment variable override runtime file ";
+            info_message += filename;
+            LoaderLogger::LogInfoMessage("", info_message);
+        } else {
 #ifdef XR_OS_WINDOWS
+            std::vector<std::string> filenames;
             ReadRuntimeDataFilesInRegistry(type, "", "ActiveRuntime", filenames);
             if (filenames.size() == 0) {
                 LoaderLogger::LogErrorMessage(
-                    "", "RuntimeManifestFile::findManifestFiles - failed to find active runtime file in registry");
+                    "", "RuntimeManifestFile::FindManifestFiles - failed to find active runtime file in registry");
                 return XR_ERROR_FILE_ACCESS_ERROR;
             }
             if (filenames.size() > 1) {
                 LoaderLogger::LogWarningMessage(
-                    "", "RuntimeManifestFile::findManifestFiles - found too many default runtime files in registry");
+                    "", "RuntimeManifestFile::FindManifestFiles - found too many default runtime files in registry");
             }
+            filename = filenames[0];
 #else
-            std::string global_rt_filename;
-            PlatformGetGlobalRuntimeFileName(XR_VERSION_MAJOR(XR_CURRENT_API_VERSION), global_rt_filename);
-            filenames.push_back(global_rt_filename);
+            if (!PlatformGetGlobalRuntimeFileName(XR_VERSION_MAJOR(XR_CURRENT_API_VERSION), filename)) {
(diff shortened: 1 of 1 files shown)


Ralith:
It's unclear what the intended behavior is in the case of multiple runtime manifests; there seems to be some suggestion that only files named `active_runtime.json` should be considered, but the prior code is clearly attempting to traverse various search paths for any `*.json` file, then processing all of them in order, with the (redundant-at-best) hardcoded path checked last.

rpavlik:
I've tagged the two most relevant folks who can address design intent here. Hopefully this can be merged soon.

daveh-lunarg:
I think there's confusion between different json file searches. Loader should only look in 1 (possibly poorly chosen) location for active_runtime.json, with an optional override via an environment variable.

The loader then looks at every .json file it finds within a half dozen paths (including XDG*) to see if it's a layer description json, but only paths ending in `openxr/<major_version>/api_layers/implicit.d` or `openxr/<major_version>/api_layers/explicit.d`. If it's opening json files in paths without one of these suffixes, it sounds like a loader bug. 

I don't mid seeing this file locking code go away, AFAICT there's no need for it.

Ralith:
If the objective is to find exactly one `active_runtime.json` file, it would be more useful to still walk the full set of search paths, but discard all but the most-specific file found. For example, this would allow users to modify the default behavior in their environment by installing a `~/.local/share/openxr/$VERSION/active_runtime.json`. Note also that different distros keep the relevant files in different locations (NixOS would probably want it in `/run/opengl-driver/share/...` for example) and referencing the XDG environment vars will make this Just Work more often, which is particularly important if the loader's being statically linked into binaries.

The pre-existing behavior is to try to load every json file floating around loose in `$XDG*` (not in the `openxr/$VERSION` subdir), and then try the hardcoded path.

daveh-lunarg:
I have a PR up to add the loader design document, which might be helpful for context.

I think ideas about how to change the active_runtime.json search behavior should go into an issue, rather than this PR discussion.

Ralith:
Opened https://github.com/KhronosGroup/OpenXR-SDK/issues/28 to discuss improving the intended behavior.

Ralith:
In the interest of reducing bugs ASAP, I've modified this PR to get the intended behavior working as documented in #27; changes to the intended behavior can be handled in a future PR.

rpavlik:
@Ralith can you take care of the CLA?

@daveh-lunarg when you return, can you re-review?

Ralith:
I think I found the right CLA thing, though it wasn't really clear where it was. Also stripped out a bit of leftover code from the PR's previous incarnation.

Ralith:
In case it wasn't clear, this fixes (among other things) behavior where the loader will try to parse loose json files in e.g. `/etc/xdg`, fail, and bail out even when a runtime is correctly installed.

src/loader/manifest_file.cpp
```diff
@@ -152,7 +152,8 @@ static void CopyIncludedPaths(bool is_directory_list, const std::string &cur_pat
         try {
             // Handle any path listings in the string (separated by the appropriate path separator)
             while (found != std::string::npos) {
-                output_path += cur_path.substr(last_found, found);
+                std::size_t length = found - last_found;
+                output_path += cur_path.substr(last_found, length);
```
rpavlik, line 156: So this is a bug fix (as is the change below), not a behavior change, right?

src/common/platform_utils.hpp
```diff
@@ -29,23 +29,6 @@
...
-    int32_t file_descriptor;
-#elif defined(XR_OS_APPLE)
-    // TBD
-#elif defined(XR_OS_WINDOWS)
-    HANDLE file_handle;
-#else
-    // TBD
-#endif
-};
-
```
rpavlik: Can you give me a TL;DR on the reason for removing this? (and related code) I'm not as well-versed in this part of the loader code, so I am having trouble evaluating it :)

It does appear to affect things described in the loader doc, specifically "7.5.3. Active Runtime File Management" - the loader wants the runtime (maybe?) to lock the runtime manifest file? Not sure how that would work when there might be more than one manifest file per runtime - e.g. Monado builds generate an "installable" one as well as a "build tree" one.

It implies that this code is for external use by runtimes, not internal use - does that match what you found?

src/loader/manifest_file.cpp
Ralith: Correct, the current incarnation of this PR is purely bugfixes and dead code removal.

src/common/platform_utils.hpp
Ralith: This commit was made before the loader doc was published, so the code appeared to be entirely dead. Reviewing that section now, I concur that it seems to be there on purpose, but I remain entirely mystified as to what purpose it could serve, and for who.

Let me know if it's a merge blocker and I'll remove it, but if it's to remain I'd like to see a concrete description of the intended use case.

rpavlik: I'd encourage splitting it into two PR's - the bug fix I can approve "right away", while the other will probably need a little WG discussion.

Ralith: Stripped it out.

rpavlik:
Thanks for splitting out that other part - this I can easily merge.

Thanks for the contribution!

PR #23, opened 2019-03-28
validation: fix deadlock in XrDestroyInstance

pH5:
`GenValidUsageNextXrDestroyInstance` locks the non-recursive `g_instance_dispatch_mutex` and then calls `GenValidUsageCleanUpMaps` under the lock, which calls `EraseAllInstanceTableMapElements`, which tries to lock `g_instance_dispatch_mutex` again, causing a deadlock.

Since `EraseAllInstanceTableMapElements` is only called from `GenValidUsageNextXrDestroyInstance`, `g_instance_dispatch_mutex` is always guaranteed to be locked, so drop the local `std::unique_lock`.

### DIFF
--- a/src/scripts/validation_layer_generator.py
+++ b/src/scripts/validation_layer_generator.py
@@ -2787,8 +2787,7 @@ class ValidationSourceOutputGenerator(AutomaticSourceOutputGenerator):
         validation_source_funcs += '\n'
         validation_source_funcs += self.outputValidationInternalProtos()
         validation_source_funcs += '// Template function to reduce duplicating the map locking, searching, and deleting.`\n'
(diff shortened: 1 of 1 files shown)


rpavlik:
oh jeepers, I totally didn't see this or your other one before submitting #24 , you must not have been in one of the channels where I warned "hey I'm changing a zillion validation things".  I'll see if I can port this across. The code is simpler post-refactoring, but still may share the underlying issue of "it's already locked", haven't dug into too far. (I did manage to shrink lock periods substantially in some cases, so there's a chance it's fixed, but not 100% sure esp. since I'm getting #25 )  

image

rpavlik (approved):
Looks good, verified to work.

rpavlik:
Merging since validation is broken without this.

PR #22, opened 2019-03-28
validation: fix creation of g_%s_info_map entries

pH5:
Create a new info entry if one doesn't already exist, not the other way around.

PR #24, opened 2019-03-28
Validation layer fixes and refactor

rpavlik:
I tried to move some things out of the generated code where possible, and reduce duplication (the locking of the maps is automatic and centralized now).

Fixed a few issues:

- Logic regarding whether to insert handle data was flipped, so the maps never got populated.
- We were using a handle to retrieve from info maps before validating that handle, meaning it would crash before it got around to sending the "invalid handle" error.
- The warning heading wasn't ever getting printed right because the case didn't match the parent class, oops.

Also went thru and ran coverage on it, removed a few large chunks of dead code from things I remember being old decisions.

I still get a segfault in a weird place during DestroySpace, but it does at least run fine until that point using a headless app.

### DIFF
--- a/src/api_layers/core_validation.cpp
+++ b/src/api_layers/core_validation.cpp
@@ -643,10 +643,7 @@ XrResult CoreValidationXrCreateSession(XrInstance instance, const XrSessionCreat
         }
         if (!got_right_graphics_binding_count) {
             std::vector<GenValidUsageXrObjectInfo> objects_info;
-            GenValidUsageXrObjectInfo handle_info = {};
-            handle_info.handle = CONVERT_HANDLE_TO_GENERIC(instance);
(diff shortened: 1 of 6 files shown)


rpavlik:
marked WIP to rebase on other PRs.

rpavlik:
OK, now rebased on #22 , investigating #23.

rpavlik:
Now rebased on both #22 and #23, with fixes confirmed for both issues. Ready to review - OK to merge after #23 and #22 are merged.

rpavlik:
OK, rebased now that I merged the other two PR's. Needs review.

daveh-lunarg (approved):
Looks good to me. I think the locks were overdone, happy so see some go away. 

Also, anything that doesn't _have_ to be generated, shouldn't be.  Good job extracting that header.

PR #26, opened 2019-04-02
hello_xr: Fix a warning

rpavlik:
My version of GCC locally pointed out that moving a return value is actually a pessimization for some reason.


daveh-lunarg (approved):
Sure.

PR #27, opened 2019-04-02
Add loader design document

daveh-lunarg:
Add the loader design document (as html).


daveh-lunarg:
With no protest noted, I'm going to go ahead and push this out for public view.

PR #30, opened 2019-04-04
Add default implementations of API functions to dispatch chains

daveh-lunarg:
The loader defines default implementations and/or terminator functions for a handful of API calls.  Two of these,  `xrResultToString` and `xrStructureTypeToString`, were not being linked into the dispatch call chains. If a runtime which didn't provide these fxns was loaded, those call chains ended with a null fxn pointer, with the expected non-optimal result.

This PR fixes the python script that generates the loader file responsible for the error.  It also refactors `LoaderInstance::CreateInstance()` to remove some unnecessary std containers that were confusing, which eventually became a pretty large re-org of that fxn. Many characters were saved in the making of this PR :)

This addresses gitlab issue: https://gitlab.khronos.org/openxr/openxr/issues/975


daveh-lunarg:
Taking commentary on the gitlab issue as approval & pushing this PR.

PR #32, opened 2019-04-05
Avoid crash when evaluating layer disable environment vars

daveh-lunarg:
The implicit layer loading code would attempt to atoi() convert a null pointer when the associated disable environment variable was not set.  Since the presence/absence of the envar is sufficient to determine intent, I removed the atoi() call entirely. Also put up an MR for the loader document to make clear that simple presence of the envar is sufficient, without specifying any particular value.

Fixes gitlab issue #976 (https://gitlab.khronos.org/openxr/openxr/issues/976)


brycehutchings (approved):

MarkY-LunarG (commented):
:+1:

PR #34, opened 2019-04-11
Fix list test output;

bjornbytes:
The list test was printing a literal `\t` instead of a tab for the `systemId` line.


daveh-lunarg (approved):
Looks good. Thanks for the PR!

PR #35, opened 2019-04-18
Add 'unknown' strings to loader's *ToString fallback functions

### LINKED ISSUE #33: xrResultToString and xrStructureTypeToString in the loader do not return the correct strings for unknown types
For an unknown structure type, xrStructureTypeToString should return XR_UNKNOWN_STRUCTURE_TYPE_*. For an unknown result code, xrResultToString should return XR_UNKNOWN_SUCCESS_*/XR_UNKNOWN_FAILURE_*.

daveh-lunarg:
The loader's fallback implementation of ResultToString and StructureToString did not correctly fill the string buffer when an unrecognized result or structure was provided.

Fixes #33

### DIFF
--- a/src/scripts/utility_source_generator.py
+++ b/src/scripts/utility_source_generator.py
@@ -244,6 +244,16 @@ class UtilitySourceOutputGenerator(AutomaticSourceOutputGenerator):
         result_to_str += self.writeIndent(indent + 1)
         result_to_str += '// Unknown result type\n'
         result_to_str += self.writeIndent(indent + 1)
+        result_to_str += 'if (XR_SUCCEEDED(result)) {\n'
+        result_to_str += self.writeIndent(indent + 2)
+        result_to_str += 'snprintf(buffer, XR_MAX_RESULT_STRING_SIZE, "XR_UNKNOWN_SUCCESS_%d", result);\n'
+        result_to_str += self.writeIndent(indent + 1)
+        result_to_str += '} else {\n'
+        result_to_str += self.writeIndent(indent + 2)
+        result_to_str += 'snprintf(buffer, XR_MAX_RESULT_STRING_SIZE, "XR_UNKNOWN_FAILURE_%d", result);\n'
+        result_to_str += self.writeIndent(indent + 1)
+        result_to_str += '}\n'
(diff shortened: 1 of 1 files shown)


src/scripts/utility_source_generator.py
```diff
@@ -244,6 +244,16 @@ def outputResultToString(self):
         result_to_str += self.writeIndent(indent + 1)
         result_to_str += '// Unknown result type\n'
         result_to_str += self.writeIndent(indent + 1)
+        result_to_str += 'if (XR_SUCCEEDED(result)) {\n'
+        result_to_str += self.writeIndent(indent + 2)
+        result_to_str += 'snprintf(buffer, XR_MAX_RESULT_STRING_SIZE, "XR_UNKNOWN_SUCCESS_%1u", (uint32_t)result);\n'
```
brycehutchings: 'result' doesn't need to be casted AFAIK (or it could be casted as int32_t like the failure case). It will still be positive as the spec requires because it is in the XR_SUCCEEDED code path.

brycehutchings: Also the 1 in "%1d" will cause it have a minimum field width of "1" (aligned to the right). Was this the intent? It doesn't seem necessary.

brycehutchings: I would recommend no casting for any of these three enums being passed into snprintf.

brycehutchings (changes requested):

src/scripts/utility_source_generator.py
daveh-lunarg: Old habits. Agreed that the casts and size 1 field width are no-ops.  (I'm sure each of them represent something that bit me somehow, some time in the last several decades, but I can't dredge up an example :)

daveh-lunarg:
I want to push this PR for correctness and to close the issue. But, in the end I'm becoming more convinced that the loader shouldn't provide these API entry points at all. (See discussion of https://gitlab.khronos.org/openxr/openxr/issues/980)

Assuming that view prevails, these util functions will eventually be removed.

brycehutchings (approved):

rpavlik:
you beat me to the review :) looks good, thanks.

PR #38, opened 2019-05-03
Remove dead file locking code

Ralith:
Split from #20

The loader design doc says:
> 7.5.3. Active Runtime File Management
> 
> Several utility functions are provided to help manage the active runtime manifest file. Of primary importance are the functions platformLockGlobalRuntimeFile and PlatformUnlockGlobalRuntimeFile which lock the file from being overwritten. These are defined in src/common/platform_utils.hpp
>
> NOTE: This is outside of the loader source to allow other items in the source folder to use these utilities.

However, there are no references to these functions whatsoever, they do not appear to be a part of any public interface, and it's unclear what use they might have.

### DIFF
--- a/src/common/platform_utils.hpp
+++ b/src/common/platform_utils.hpp
@@ -27,23 +27,6 @@
 #include <iostream>
 #endif
 
-// Structure used to track the global runtime file information
-struct GlobalRuntimeFile {
-    bool valid;
-    bool locked;
-    std::string file_name;
-    std::string runtime_name;
-#if defined(XR_OS_LINUX)
-    int32_t file_descriptor;
-#elif defined(XR_OS_APPLE)
-    // TBD
-#elif defined(XR_OS_WINDOWS)
-    HANDLE file_handle;
-#else
-    // TBD
-#endif
-};
-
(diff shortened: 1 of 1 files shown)


rpavlik:
Thanks for the patch - I've raised an internal issue to look into this.

daveh-lunarg:
I've had a chance to discuss with @MarkY-LunarG the intent behind this file locking code. Originally intended as a mechanism to prevent multiple runtimes from being loaded simultaneously, but later removed (or _calls_ removed, leaving this as dead code) as not super effective for the task and WG discussion may have raised some scenarios where multiple runtimes might be allowed/useful (? dunno - before my time).

Just on principle of not allowing unused code to rot within the repo, I'm happy to see this code excised. I'll remove the reference in the loader design doc to agree.

brycehutchings:
Happy to see this go because it doesn't even work on Windows where the registry key can't be locked like a file can.

daveh-lunarg:
Accompanying PR to remove the locking discussion from the loader design document is PR #50

daveh-lunarg (approved):
Looks good to me.  Builds/tests with no adverse affect on Windows for me.

PR #39, opened 2019-05-06
loader: Default to static loader only on windows

ChristophHaag:
On Linux we want to follow the Vulkan model of a system wide dynamic loader.
Appplications are still free to link to a static copy of the loader if they wish.


rpavlik:
Thanks!

PR #40, opened 2019-05-07
Update BUILDING.md with some linux pre-requisites

daveh-lunarg:
Just stood up a vanilla Ubuntu 19.04, these are the minimal packages I had to install to get the build working.

@rpavlik, maybe you know more about xxf86vm? Installing it alone didn't satisfy cmake's findpackage, but then installing xrandr resolved the findpackage for xrandr and xxf86vm. So, maybe xxf86vm isn't really needed?  I'm assuming no harm in listing it.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -46,12 +46,20 @@ Open the build\win32\OPENXR.sln in the Visual Studio to build the samples.
 
 ## Linux
 
-The following packages are required, at minimum.
+The following set of packages provides all required libs for building for xlib or xcb with OpenGL and Vulkan support.
+ - build-essential
  - cmake (of _somewhat_ recent vintage, 3.10+ known working)
- - libx11-dev
- - libxxf86vm-dev
+ - libgl1-mesa-dev
+ - libvulkan-dev
+ - libx11-xcb-dev
+ - libxcb-dri2-0-dev
+ - libxcb-glx0-dev
+ - libxcb-icccm4-dev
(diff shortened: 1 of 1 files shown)


rpavlik:
Well, I think it's OK to leave that (though I don't think it's required if you are building wayland-only.)  Since mid-day yesterday we've been building internally on CI using this dockerfile: https://github.com/KhronosGroup/DockerContainers/blob/master/Dockerfile.openxr-sdk  which shows a list of packages. It's not an absolute minimum list because it's set up to build everything possible, but it should give you an idea. (freeglut isn't required, you're mostly just getting the mesa dev files with that as a dep.)

EDIT: I think the list of packages required for xlib or XCB builds with both OpenGL and Vulkan is:

```
    build-essential \
    cmake \
    libgl1-mesa-dev \
    libvulkan-dev \
    libx11-xcb-dev \
    libxcb-dri2-0-dev \
    libxcb-glx0-dev \
    libxcb-icccm4-dev \
    libxcb-keysyms1-dev \
    libxcb-randr0-dev \
    libxrandr-dev \
    libxxf86vm-dev \
    mesa-common-dev
```

Drop glx, libgl1-mesa-dev, and maybe mesa-common-dev if you don't want OpenGL, or drop libvulkan-dev if you don't want vulkan.

Drop anything starting with an `x` if you don't care about X11 support (want wayland instead).

Extra packages for wayland:

- libegl1-mesa-dev
- libwayland-dev
- wayland-protocols

PR #43, opened 2019-05-10
make mainfest file more compatible

zheqiMicrosoft:
1. Adding #include\<sstream\> so manifest_file.cpp can work with jsoncpp library without having to almagate the rouces together (calling src/external/jsoncpp/amalgamate.py). This is useful with chroumium where jsoncpp library is used as not almagated.
  
2. Changing Json:::ValueIterator func_it-> to Json:::ValueIterator (*func_it). so it will be compatible to older version of jsoncpp library.(like the one used by chroumium now)

### DIFF
--- a/src/loader/manifest_file.cpp
+++ b/src/loader/manifest_file.cpp
@@ -24,6 +24,7 @@
 #include <cstring>
 #include <fstream>
 #include <iostream>
+#include <sstream>
 #include <stdexcept>
 
 #include "filesystem_utils.hpp"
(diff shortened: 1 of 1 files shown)


brycehutchings:
[EDIT: Disregard! I misread what repo this was in]

yl-msft:
Tested this change on openxr samples and it seems working fine.

yl-msft (approved):

rpavlik (approved):

PR #44, opened 2019-05-12
Remove pkgconfig dependency from xlib backend

jherico:
The `vcpkg` CI environment does not include `pkg-config`, so the `presentation.cmake` file fails.   However, the bulk of the `presentation.cmake` (and the use of `pkg-config` in particular) sets link libraries that are only used by the test code.  

This PR moves all of the references to `pkg-config` inside of an `if (BUILD_TESTS) ... endif()` block, while migrating the `add_definitions` out to keep that logic.

### DIFF
--- a/src/cmake/presentation.cmake
+++ b/src/cmake/presentation.cmake
@@ -12,21 +12,22 @@ endif()
 
 message(STATUS "Using presentation backend: ${PRESENTATION_BACKEND}")
 
-find_package(PkgConfig REQUIRED)
 
 if( PRESENTATION_BACKEND MATCHES "xlib" )
-    pkg_search_module(X11 REQUIRED x11)
-    pkg_search_module(XXF86VM REQUIRED xxf86vm)
-    pkg_search_module(XRANDR REQUIRED xrandr)
+    find_package(X11 REQUIRED)
+    if ((NOT X11_Xxf86vm_LIB) OR (NOT X11_Xrandr_LIB))
+        message(FATAL_ERROR "OpenXR xlib backend requires Xxf86vm and Xrandr")
+    endif()
 
     add_definitions( -DSUPPORT_X )
     add_definitions( -DOS_LINUX_XLIB )
     set( XLIB_LIBRARIES
             ${X11_LIBRARIES}
-            ${XXF86VM_LIBRARIES}
-            ${XRANDR_LIBRARIES} )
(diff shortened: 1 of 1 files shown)


rpavlik:
It does build for me here, although I am suspicious that it may not build on a system that actually lacks the X11 headers, since many files include `xr_dependencies.h` which include that.

Ideally, we'd go to a proper CMake module, possibly using pkg-config as hints, as recommended by the CMake docs. IIRC, I looked into this a while ago and found a lack of such ones pre-written for XCB.

jherico:
The vcpkg CI Linux servers clearly have libx11-dev, since I know there are at least a few packages like SDL and GLFW that wouldn't compile otherwise.  They just don't have `PkgConfig`....

OK... so looking at how GLFW deals with this provided some interesting insight... For xrandr and xf86vm, you don't need to use pkg-config...  the CMake X11 module will actually populate what you need.  If I do `find_package(X11)` and dump the variables I see:

    X11_Xrandr_FOUND=TRUE
    X11_Xrandr_INCLUDE_PATH=/usr/include
    X11_Xrandr_LIB=/usr/lib/x86_64-linux-gnu/libXrandr.so
    X11_Xxf86vm_LIB=/usr/lib/x86_64-linux-gnu/libXxf86vm.so

That completely covers the xlib backend.  GLFW also gets away with not using PkgConfig.  However it does this by forward declaring everything rather than using xcb headers, and then using [explicit loading of the shared library and function](https://github.com/glfw/glfw/blob/a337c568486025d22443535b7c047a563bc34373/src/x11_init.c#L668)

Wayland, it seems, simply can't be done without `PkgConfig`.

At any rate, the fact that the X11 module contains everything needed for the default xlib backend means I could alter this PR to make a smaller change.  Instead of enclosing everything in the `BUILD_TESTS` conditional, I could modify the xlib backend to use the `FindPackage(X11)` logic, and then move the `find_package(PkgConfig REQUIRED)` into the `xcb` and `wayland` paths.  

Would that be preferable?

rpavlik:
So does this solve your problem for now? I'll eventually get to replacing the other two with a cmake module, if somebody out there doesn't beat me to it, but if this unblocks you, I'm happy to merge it now.

jherico:
Yes, I believe this is everything I need for now to get a vcpkg build without patches.

PR #46, opened 2019-05-14
Support building with embedded python

jherico:
The CMake logic uses the environment variable `PYTHONPATH` to allow the `src_genxr.py` and `genxr.py` scripts to execute using local modules like `reg`.  

However, vcpkg will use an embedded version of Python if a conventional version is not found on the host system, and the embedded version of Python will not respect the `PYTHONPATH` env variable.  

This PR modifies the scripts to internally modify the `sys.path` array with the required paths, defined here as relative paths to the root of the repository.

This removes the need for defining PYTHONPATH in the CMake files, but PR hasn't done so.

### DIFF
--- a/specification/scripts/genxr.py
+++ b/specification/scripts/genxr.py
@@ -15,10 +15,13 @@
 # limitations under the License.
 
 import argparse
+import os
 import re
 import sys
 import time
 import xml.etree.ElementTree as etree
+
+sys.path.append(os.path.abspath(os.path.dirname(__file__)))
 
 from cgenerator import CGeneratorOptions, COutputGenerator
 from docgenerator import DocGeneratorOptions, DocOutputGenerator
--- a/src/scripts/src_genxr.py
+++ b/src/scripts/src_genxr.py
@@ -14,7 +14,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import argparse, cProfile, pdb, string, sys, time
+import argparse, cProfile, os, pdb, string, sys, time
+
+base_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..'))
+sys.path.append(os.path.join(base_dir, 'src', 'scripts'))
+sys.path.append(os.path.join(base_dir, 'specification', 'scripts'))
+
 from reg import *
 from generator import write
 from cgenerator import CGeneratorOptions, COutputGenerator


specification/scripts/genxr.py
```diff
@@ -15,11 +15,17 @@
...
 
 import argparse
+import os
 import re
 import sys
 import time
 import xml.etree.ElementTree as etree
 
+base_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..'))
+sys.path.append(os.path.join(base_dir, 'src', 'scripts'))
```
rpavlik: So the scripts in `specification/scripts` only depend on things in that directory. It's the ones in `src/scripts` that depend on files there as well as in `specification/scripts` - the `src` scripts build on the specification ones. So, I think this line is unnecessary.

(Is any of this necessary in this directory? I regularly run "specification" scripts  from another directory without setting python path. The stuff over in `src/` is definitely necessary though.)

rpavlik (changes requested):
If it's easier, I think it's OK to use pathlib if you like - I have already used it elsewhere in the scripts. On the other hand, if this works, then great :)  Just one little nitpick noted.

specification/scripts/genxr.py
jherico: I think you're correct that the `src/scripts` path isn't required to execute `genxy.py`.  I've updated the PR.

However, adding the `specification/scripts` folder is definitely necessary for working with embedded python 3.7.3 and adding _both_ directories is necessary for the `src_genxy.py` script.  

I can create a simple file `foo.py` in the root of the repository with the following contents

    #!/usr/bin/python3
    import sys
    print(sys.path)

If I execute it with my installed Python I get the following output 

```
$ "/c/Program Files (x86)/Microsoft Visual Studio/Shared/Python36_64/python3" ~/git/OpenXR-SDK/foo.py
['C:\\Users\\bdavi\\git\\OpenXR-SDK', 'C:\\Program Files (x86)\\Microsoft Visual Studio\\Shared\\Python36_64\\python36.zip', 'C:\\Program Files (x86)\\Microsoft Visual Studio\\Shared\\Python36_64\\DLLs', 'C:\\Program Files (x86)\\Microsoft Visual Studio\\Shared\\Python36_64\\lib', 'C:\\Program Files (x86)\\Microsoft Visual Studio\\Shared\\Python36_64', 'C:\\Users\\bdavi\\AppData\\Roaming\\Python\\Python36\\site-packages', 'C:\\Program Files (x86)\\Microsoft Visual Studio\\Shared\\Python36_64\\lib\\site-packages']
```

The location of the script (`C:\\Users\\bdavi\\git\\OpenXR-SDK`) is automatically prepended to `sys.path` by Python, so the `import reg` statement works.

Similarly if I execute it using the _current_ embedded Python (3.5.4) we see the same thing

```
$ ~/git/vcpkg/downloads/tools/python/python3/python.exe ~/git/OpenXR-SDK/foo.py
['C:\\Users\\bdavi\\git\\OpenXR-SDK', 'C:\\Users\\bdavi\\git\\vcpkg\\downloads\\tools\\python\\python3\\python35.zip', 'C:\\Users\\bdavi\\git\\vcpkg\\downloads\\tools\\python\\python3\\DLLs', 'C:\\Users\\bdavi\\git\\vcpkg\\downloads\\tools\\python\\python3\\lib', 'C:\\Users\\bdavi\\git\\vcpkg\\downloads\\tools\\python\\python3']
```

However, that build fails because of a bug in Python3 < 3.6 where a WindowsPath is not treated as a pathlike.  Hence vcpkg has an open PR to switch to the latest Python embedded version... 3.7.3  ... when I execute the same script with that version...

```
$ /f/Downloads/python-3.7.3-embed-win32/python.exe ~/git/OpenXR-SDK/foo.py
['F:\\Downloads\\python-3.7.3-embed-win32\\python37.zip', 'F:\\Downloads\\python-3.7.3-embed-win32']
```

Not only does it ignore `PYTHONPATH`, it doesn't even add the script directory to `sys.path`.  I don't know if the change in behavior between 3.5.4 and 3.7.3 is considered a bug or a fix.

rpavlik: Ahh, interesting that it doesn't even add the script dir to the `sys.path`. It sounds like a bug, but who knows, it might be a security fix of some sort. OK, works for me.

rpavlik (approved):

rpavlik:
Thanks for your contribution!

PR #48, opened 2019-05-15
Install layers and pkg-config file on Linux

ChristophHaag:
These changes are inspired by the Vulkan loader build.


daveh-lunarg (approved):
Looks reasonable to me.

rpavlik (approved):

PR #51, opened 2019-05-18
Idiomatic Linux active_runtime.json search logic

### LINKED ISSUE #28: Linux active runtime search
The draft loader design document (#27) states:

> On Linux, the active runtime information is contained inside a JSON-formatted file located under the folder:
>
>     /usr/local/share/openxr/<major_api_version>`

I propose a different strategy: select the first `openxr/$VERSION/active_runtime.json` found by searching first `$XDG_CONFIG_HOME`, then the elements of `$XDG_CONFIG_DIRS`, in order. If an environment variable is not set or the process is SUID, the freedesktop.org fallback values are used, unless a more appropriate fallback is specified by the target environment.

This has a number of advantages:

- Flexibility: a user who wishes to configure a default runtime other than the one configured globally can do so by placing a file in `$XDG_CONFIG_HOME/openxr/$VERSION/active_runtime.json` without requiring elevated privileges.
- Convention: `/usr/local` is not a standard prefix for packages installed through the system's package manager, and runtime packages targeting that location are unlikely to be supported by major distributions. It is also not a standard prefix for administrator-controlled configuration files associated with system infrastructure such as a package-manager-installed OpenXR runtime or loader.
- Portability: Not all distributions implement the filesystem hierarchy standard. For example, `/usr/local` does not exist on NixOS, and indeed no NixOS packages may write to `/usr` at all. Such environments provide XDG environment variables identifying the correct locations, allowing precompiled binaries to Just Work even if they were not specifically targeted to the environment. Such environments can also override the fallback paths in their packaging of the OpenXR loader.

Ralith:
Based on #38. Fixes #28.

This updates the search logic to find the first active_runtime.json in standard config file locations, allowing for more graceful integration with Linux distribution tooling and user/administrator expectations.

I opted not to reuse `ReadDataFilesInSearchPaths` due to its complexity and conflation of config/data search paths.

If this is to be merged, the loader design doc will also need to be updated. I'm not sure where to find the source files for that.

### DIFF
--- a/src/common/platform_utils.hpp
+++ b/src/common/platform_utils.hpp
@@ -59,17 +59,6 @@ static inline void PlatformUtilsFreeEnv(char* val) {
     (void)val;
 }
 
-// Prefix for the Linux/Apple global runtime JSON file name
-static const std::string rt_dir_prefix = "/usr/local/share/openxr/";
-static const std::string rt_filename = "/active_runtime.json";
-
-static inline bool PlatformGetGlobalRuntimeFileName(uint16_t major_version, std::string& file_name) {
-    file_name = rt_dir_prefix;
-    file_name += std::to_string(major_version);
-    file_name += rt_filename;
-    return true;
-}
-
 #elif defined(XR_OS_APPLE)
 
 static inline char *PlatformUtilsGetEnv(const char *name) { return getenv(name); }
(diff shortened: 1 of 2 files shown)


rpavlik:
I'm checking to see if there's any paperwork/admin stuff we need to do to put the loader docs up. I suspect not, since it's not normative. Think I'll just put the source for them in this repo, since they don't really affect the spec itself. (that make sense?)

daveh-lunarg (approved):
LGTM, fixes the issue.

daveh-lunarg:
@rpavlik, I think we're good to go to push this - any objection?

Ralith:
Updated to avoid assigning a null `char*` to a `std::string` in `GetXDGEnv`.

daveh-lunarg:
Hearing no dissent, pushing this.

PR #55, opened 2019-06-12
loader: purge std::map bracket operations that might do inadvertent insertions

### LINKED ISSUE #53: Verify or fix std::map bracket operator insertions
There are a few instances in the loader code where a std::map is accessed via bracket operator
```
foo = map[key]
```
where it appears uncertain whether the key already exists in the map.  These should be changed to or guarded by .find() operations, to prevent an inadvertent insertion of the key into the map.

Tracks WG gitlab issue # [943](https://gitlab.khronos.org/openxr/openxr/issues/943)

daveh-lunarg:
There were several instances where a `map[key]` operations appear on the RHS of an assignment, i.e. as a map lookup.  If the `key` is not present in the map, it will be inserted with a null value by the bracket operation, which is almost certainly not the intent.

Replaced all occurrences with `.find()` or guarded with `.count()` to prevent unintended insertions.

Resolves #53

### DIFF
--- a/src/api_layers/api_dump.cpp
+++ b/src/api_layers/api_dump.cpp
@@ -573,8 +573,17 @@ XrResult ApiDumpLayerXrDestroyInstance(XrInstance instance) {
     ApiDumpLayerRecordContent(contents);
 
     std::unique_lock<std::mutex> mlock(g_instance_dispatch_mutex);
-    XrGeneratedDispatchTable *next_dispatch = g_instance_dispatch_map[instance];
+    XrGeneratedDispatchTable *next_dispatch = nullptr;
+    auto map_iter = g_instance_dispatch_map.find(instance);
+    if (map_iter != g_instance_dispatch_map.end()) {
+        next_dispatch = map_iter->second;
+    }
     mlock.unlock();
+
+    if (nullptr == next_dispatch) {
+        return XR_ERROR_HANDLE_INVALID;
+    }
+
     next_dispatch->DestroyInstance(instance);
     ApiDumpCleanUpMapsForTable(next_dispatch);
 
--- a/src/loader/loader_core.cpp
+++ b/src/loader/loader_core.cpp
@@ -251,7 +251,7 @@ LOADER_EXPORT XRAPI_ATTR XrResult XRAPI_CALL xrCreateInstance(const XrInstanceCr
             }
         }
 
-        if (XR_SUCCESS != result) {
+        if (XR_FAILED(result)) {
             if (runtime_loaded) {
                 RuntimeInterface::UnloadRuntime("xrCreateInstance");
             }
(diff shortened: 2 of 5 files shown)


src/loader/loader_core.cpp
```diff
@@ -267,10 +267,14 @@ LOADER_EXPORT XRAPI_ATTR XrResult XRAPI_CALL xrCreateInstance(const XrInstanceCr
...
             *instance = created_instance;
 
-            LoaderInstance *loader_instance;
+            LoaderInstance *loader_instance = nullptr;
             {
                 std::unique_lock<std::mutex> lock(g_instance_mutex);
-                loader_instance = g_instance_map[created_instance];
+                auto map_iter = g_instance_map.find(created_instance);
+                if (map_iter == g_instance_map.end()) {
+                    return XR_ERROR_VALIDATION_FAILURE;
```
brycehutchings: If this error happened, it looks like it is a bug in the loader because it should have put it in the map previously. I don't believe there is a good error code available though (like an XR_ERROR_LOADER_FAILURE error). Perhaps just adding logging (LoaderLogger::LogErrorMessage) is sufficient to make this debuggable if it ever happened.

src/api_layers/api_dump.cpp
```diff
@@ -573,8 +573,17 @@ XrResult ApiDumpLayerXrDestroyInstance(XrInstance instance) {
...
-    XrGeneratedDispatchTable *next_dispatch = g_instance_dispatch_map[instance];
+    XrGeneratedDispatchTable *next_dispatch = nullptr;
+    auto map_iter = g_instance_dispatch_map.find(instance);
+    if (map_iter != g_instance_dispatch_map.end()) {
+        next_dispatch = map_iter->second;
+    }
     mlock.unlock();
+
+    if (nullptr == next_dispatch) {
+        return XR_ERROR_VALIDATION_FAILURE;
```
brycehutchings: Shouldn't this be XR_ERROR_HANDLE_INVALID?

src/loader/runtime_interface.cpp
```diff
@@ -306,8 +311,11 @@ XrResult RuntimeInterface::DestroyInstance(XrInstance instance) {
         if (XR_NULL_HANDLE != instance) {
             // Destroy the dispatch table for this instance first
             std::unique_lock<std::mutex> mlock(_dispatch_table_mutex);
-            XrGeneratedDispatchTable* table = _dispatch_table_map[instance];
-            _dispatch_table_map.erase(instance);
+            XrGeneratedDispatchTable* table = nullptr;
+            if (0 != _dispatch_table_map.count(instance)) {
```
brycehutchings: this use of count followed by [] can be combined into a single find() call because you can check if it exists (!= end) and erase has an overload for taking an iterator.

brycehutchings (dismissed):
The thing I wasn't sure about was when XR_ERROR_VALIDATION_ERROR is returned versus XR_ERROR_HANDLE_INVALID.

src/api_layers/api_dump.cpp
daveh-lunarg: Yep, agreed.

src/loader/loader_core.cpp
daveh-lunarg: Didn't really parse the logic here, just doing search and destroy on RHS bracket operators. With a little more thought I don't see the need for a check here at all - if the create immediately above succeeded then no need to check for a missing map entry.

daveh-lunarg:
Updated with @brycehutchings suggestions, and pushed.

PR #56, opened 2019-06-12
Making filesystem_util.cc UNICODE Compatible

zheqiMicrosoft:
make filesystem_util unicode compatible when USE_EXPERIMENTAL_FS=0 and USE_FINAL_FS=0


brycehutchings:
We have started the process of fixing the loader to use the "W" Win32 functions for proper Unicode support. This pull request may conflict/duplicate that work. I think for now in GitHub, switching to explicitly using the "A" functions will fix building with the UNICODE flag enabled (the loader is doing this already in some places, like RegOpenKeyExA in manifest_file.cpp in the loader and in the private OpenXR repo this has already been fixed to use RegOpenKeyExW but not yet brought over to this GitHub repo). I think fixing the build issue by explicitly using the ANSI "A" functions here in GitHub is the best fix to do here before we bring over the fixed version which uses the wide "W" functions so minimize merge pain. 

Does that work for you?

zheqiMicrosoft:
I think that works. Thanks!

brycehutchings (approved):

rpavlik (approved):

PR #57, opened 2019-06-14
Make it possible to bypass marco that check which filesystem to use

zheqiMicrosoft:
During some special circumstances(like chromium build system). The logic that determines which system to use for filesystem_utils.cpp may be false. By adding these two lines, we can directly control which library to use and bypass those logic.


daveh-lunarg (approved):
Looks good to me

Issue #54, opened 2019-06-12
Loader breaks on unexpected structure


rpavlik:
I think we should be able to pass any kind of struct in to any function on the next chain, at least through the loader, but I can't:

```
Error [GENERAL | xrCreateInstance | OpenXR-Loader] : VUID-XrInstanceCreateInfo-next-next: unexpected struct in 'next' chain.
Error [GENERAL | xrCreateInstance | OpenXR-Loader] : VUID-xrCreateInstance-info-parameter: something wrong with XrInstanceCreateInfo contents
Error [GENERAL | xrCreateInstance | OpenXR-Loader] : LoaderInstance::CreateInstance chained CreateInstance call failed
```

This was prompted by this test: https://gitlab.freedesktop.org/monado/utilities/xr-system-tests/blob/master/tests/functions/xrCreateInstance.cpp#L53  which throws an `XrActionCreateInfo` in the `XrInstanceCreateInfo` chain.

daveh-lunarg:
Assigned myself. Buried on another project for a couple weeks, plan to look at this in early July.

rpavlik:
oh go figure - looks mostly like an easy fix - there's a ValidateInstanceCreateInfoNextChain which doesn't really need to exist. Not seeing any associated ones in generated code for the rest - guess I'll have to try messing around with next chains in other functions.

daveh-lunarg:
Closing - resolved by PR #62.

PR #62, opened 2019-06-17
loader: Don't limit contents of XrInstanceCreateInfo next chain.

rpavlik:
At least a partial fix for #54 - couldn't find any other places where similar logic remained, with a simple search, but may have missed some.

### DIFF
--- a/src/loader/loader_core.cpp
+++ b/src/loader/loader_core.cpp
@@ -355,19 +355,6 @@ LOADER_EXPORT XRAPI_ATTR XrResult XRAPI_CALL xrDestroyInstance(XrInstance instan
 
 // ---- Core 0.1 manual loader terminator functions
 
-// Validate that the XrInstanceCreateInfo 'next' chain is valid.
-static bool ValidateInstanceCreateInfoNextChain(LoaderInstance *loader_instance, const XrInstanceCreateInfo *info) {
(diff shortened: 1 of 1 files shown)


daveh-lunarg (changes requested):
I think the `ValidateInstanceCreateInfoNextChain` fxn is correct, though incomplete, based on this Valid Usage (Implicit) statement for the XrInstanceCreateInfo struct:
```
Each next member of any structure (including this one) in the next chain must be either NULL
or a pointer to a valid instance of XrDebugUtilsMessengerCreateInfoEXT or
XrInstanceCreateInfoAndroidKHR
```
It looks to me like it should remain, and just add the `XrInstanceCreateInfoAndroidKHR` as an allowed type.

rpavlik:
unfortunately the generated valid usage there is probably wrong since it conflicts with the explicit prose we have elsewhere - see internal issue 1051

daveh-lunarg:
Ok, I'm looking at a 0.9 spec and there's no reason to prohibit unknown structs. It makes me a little uneasy to change code to match what the spec _should have_ said, but it's early days in XR land.

daveh-lunarg (approved):

rpavlik:
thanks for review! (In any case, it seems a little weird that the loader would validate just this chain, when it's not essential to its job)

PR #63, opened 2019-06-18
Allow certain hardware time to initialize.

Szalacinski:
Would not work previously using Monado with the Oculus CV1.  There are some magic commands that need to be sent to the CV1 over USB to turn the display on and have X recognize it.  This delay allows the device to startup before hello_xr attempts to access it.  Before this addition, hello_xr would complain that no non-desktop display devices were found, and now it works out of the box with OpenHMD 0.3.0-rc1 and the master branch of Monado.


Szalacinski:
I'm wondering if this would be better off in the OpenHMD driver, because developers probably don't need to be thinking about these kinds of timing issues when working with a high-level API.  Let me know what you think.

brycehutchings:
Yeah, this is a runtime bug and needs to be fixed in the runtime. Applications should not include a delay.

Szalacinski:
Ok. I'll take this to OpenHMD.

PR #65, opened 2019-06-24
fix minor substr error

zheqiMicrosoft:
There is a minor bug with substring usage. Remove the "- 1" that's unnecessary.

### DIFF
--- a/src/common/filesystem_utils.cpp
+++ b/src/common/filesystem_utils.cpp
@@ -259,7 +259,7 @@ bool FileSysUtilsGetParentPath(const std::string& file_path, std::string& parent
         std::string full_path;
         if (FileSysUtilsGetAbsolutePath(file_path, full_path)) {
             std::string::size_type lastSeperator = full_path.find_last_of(DIRECTORY_SYMBOL);
-            parent_path = (lastSeperator == 0) ? full_path : full_path.substr(0, lastSeperator - 1);
+            parent_path = (lastSeperator == 0) ? full_path : full_path.substr(0, lastSeperator);
             return true;
         }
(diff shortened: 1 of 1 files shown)


yl-msft (approved):

rpavlik:
So this was put in by a Linux dev, IIRC. Only risk I see is that it might result in a doubled-up path character (/ or \) in its usage, which I wouldn't expect to be harmful, but not sure. What is the actual bug being corrected here?

zheqiMicrosoft:
> hat is the actual bug being corrected here?

zheqiMicrosoft:
The question I believe is that in windows when the path is "D:\chromium2\src\out\Debug\loader.dll" the parent path will be "D:\chromium2\src\out\Debu" instead of ""D:\chromium2\src\out\Debug".

patrto:
> So this was put in by a Linux dev, IIRC. Only risk I see is that it might result in a doubled-up path character (/ or ) in its usage, which I wouldn't expect to be harmful, but not sure. What is the actual bug being corrected here?

In Windows, the -1 is truncating the parent path of the runtime:
C:\Windows\System32\OpenXRRuntime.dll -> C:\Windows\System3\OpenXRRuntime.dll

Is there a Linux case that breaks without the -1?

yl-msft:
> Is there a Linux case that breaks without the -1?

isn't this a windows only code path?  not sure how can Linux be affected by this change.

rpavlik:
Sounds good to me. I just remember that line being added by (iirc) a primarily-Linux dev. I'm not as familiar with that piece of the loader.

PR #66, opened 2019-06-25
build: install the layers libraries on linux

ChristophHaag:
My previous PR #48 apparently forgot to install the layers themselves.


rpavlik:
No idea if this is the expected location for these, but it's more correct that what's in there :)

PR #69, opened 2019-06-25
Simplify loader

rpavlik:
The main thing I did here is the same basic cleanup I performed on the validation layer a while ago: replaced duplicated "lock, look up in map, unlock" logic with classes that do that all in one call.

The manual new and delete made me nervous so I switched nearly all of them to unique_ptrs now too.

It also changes a few XR_ERROR_VALIDATION_FAILURE into XR_ERROR_HANDLE_INVALID for consistency.

Handful of other changes too. De-duplicates and makes it easier at least for me to read and follow, hopefully for others too.

Commit messages should be useful.

### DIFF
--- a/src/loader/loader_instance.hpp
+++ b/src/loader/loader_instance.hpp
@@ -24,9 +24,11 @@
 
 #include "loader_platform.hpp"
 #include "platform_utils.hpp"
+#include "extra_algorithms.h"
 #include "runtime_interface.hpp"
 #include "api_layer_interface.hpp"
 #include "xr_generated_dispatch_table.h"
+
 class LoaderInstance;
 
 typedef std::unique_lock<std::mutex> UniqueLock;
@@ -56,21 +58,6 @@ class HandleLoaderMap {
     map_t instance_map_;
     std::mutex mutex_;
 };
-
-/// Like std::remove_if, except it works on associative containers and it actually removes this.
-///
-/// The iterator stuff in here is subtle - .erase() invalidates only that iterator, but it returns a non-invalidated iterator to the
-/// next valid element which we can use instead of incrementing.
-template <typename T, typename Pred>
-inline void map_erase_if(T& container, Pred&& predicate) {
-    for (auto it = container.begin(); it != container.end();) {
-        if (predicate(*it)) {
-            it = container.erase(it);
-        } else {
-            ++it;
-        }
-    }
-}
 
 class LoaderInstance {
    public:
--- a/src/loader/loader_instance.hpp
+++ b/src/loader/loader_instance.hpp
@@ -44,7 +44,8 @@ class HandleLoaderMap {
     LoaderInstance* Get(HandleType handle);
 
     /// Insert an info for the supplied handle.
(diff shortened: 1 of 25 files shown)


rpavlik:
ci error is 32-bit reinterpret_cast of handles, relatively easy fix.

rpavlik:
OK, that should have fixed it in such a way that it won't break again (and neither will other places in the source since I went through and unified things.) I reviewed every `std::hex` and every `reinterpret_cast` in the src directory and the scripts directory, changing all those that did something done in the new xr_utils.h header to just use that directly.

src/loader/loader_instance.hpp
```diff
@@ -27,14 +27,59 @@
...
+};
+
+/// Like std::remove_if, except it works on associative containers and it actually removes this.
+///
+/// The iterator stuff in here is subtle - .erase() invalidates only that iterator, so we must advance first.
+template <typename T, typename Pred>
+inline void map_erase_if(T& container, Pred&& predicate) {
+    for (auto it = container.begin(); it != container.end();) {
+        if (predicate(*it)) {
+            auto here = it;
```
brycehutchings: I think you can avoid this iterator dance by using the return value from erase which is an iterator pointing to the next element.

src/loader/loader_instance.hpp
```diff
@@ -61,3 +106,56 @@ class LoaderInstance {
...
+template <typename HandleType>
+inline XrResult HandleLoaderMap<HandleType>::Insert(HandleType handle, LoaderInstance& loader) {
+    if (handle == XR_NULL_HANDLE) {
+        // Internal error in loader or runtime.
+        return XR_ERROR_RUNTIME_FAILURE;
+    }
+    UniqueLock lock(mutex_);
+    auto entry_returned = instance_map_.find(handle);
+    if (entry_returned != instance_map_.end()) {
+        // Internal error in loader or runtime.
```
brycehutchings, line 123: I would be careful here. The loader isn't great at managing handle lifetimes. For example, if an app calls xrDestroyActionSet then all of the child actions are also destroyed, but the loader won't clean up those actions. So a runtime may return newly created action handles that reuse existing handle values, which is legit, but it will cause this error to happen.

brycehutchings (changes requested):

src/loader/loader_instance.hpp
rpavlik: Ah, cool - looked it up, and looks like the overloads that take an iterator do actually return one. Cool, will do.

src/loader/loader_instance.hpp
rpavlik: Ah, good point. This was just a transplant of the logic that was spread out all over before, but I can address that.

brycehutchings (approved):

rpavlik:
OK, got 32-bit compilation working locally, which inexplicably revealed an unresolved symbol error that didn't show up when building 64-bit. Hopefully this get the job done.

rpavlik:
OK, got this tested to at least build on Windows. Bizarre that clang was fine building it on Linux.

PR #70, opened 2019-06-26
Make Loader Exception Free

zheqiMicrosoft:
Try to make exception free by doing the following:
1. For Exceptions are thrown and caught directly by the function that throws. These instances are changed to early return with a corresponding XrResult and logging, if any.
2. For Throws that are the result of a catch(..) with some “unknown error” or “bad alloc” logging. Remove these instances
3. Remove all catches directly.
4. Change JSON_USE_EXCEPTION to be defined as 0 so jsoncpp will not throw exception.

Note: This PR does not touch throws and catches in all tests, core_validation and xr_generated_core_validation. core_validation and xr_generated_core_validation are intended to be used by application developers and not part of the loader.

### DIFF
--- a/src/api_layers/api_dump.cpp
+++ b/src/api_layers/api_dump.cpp
@@ -59,133 +59,125 @@ static std::mutex g_record_mutex = {};
 
 // HTML utilities
 bool ApiDumpLayerWriteHtmlHeader(void) {
-    try {
-        std::unique_lock<std::mutex> mlock(g_record_mutex);
-        std::ofstream html_file;
-        html_file.open(g_record_info.file_name, std::ios::out);
-        html_file << "<!doctype html>\n"
-                     "<html>\n"
-                     "    <head>\n"
-                     "        <title>OpenXR API Dump</title>\n"
-                     "        <style type='text/css'>\n"
-                     "        html {\n"
-                     "            background-color: #0b1e48;\n"
-                     "            background-image: url('https://vulkan.lunarg.com/img/bg-starfield.jpg');\n"
-                     "            background-position: center;\n"
-                     "            -webkit-background-size: cover;\n"
-                     "            -moz-background-size: cover;\n"
-                     "            -o-background-size: cover;\n"
-                     "            background-size: cover;\n"
-                     "            background-attachment: fixed;\n"
-                     "            background-repeat: no-repeat;\n"
-                     "            height: 100%;\n"
-                     "        }\n"
-                     "        #header {\n"
-                     "            z-index: -1;\n"
-                     "        }\n"
-                     "        #header>img {\n"
-                     "            position: absolute;\n"
-                     "            width: 160px;\n"
-                     "            margin-left: -280px;\n"
-                     "            top: -10px;\n"
-                     "            left: 50%;\n"
-                     "        }\n"
-                     "        #header>h1 {\n"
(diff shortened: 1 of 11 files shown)


zheqiMicrosoft:
@daveh-lunarg Here is my PR for making loader exception free. Thanks!

rpavlik:
Looks like you also changed the ApiDump layer as well, which is probably fine - haven't gotten thru it - but is above and beyond doing just the loader. :)

src/api_layers/api_dump.cpp
```diff
@@ -573,17 +561,8 @@ XrResult ApiDumpLayerXrDestroyInstance(XrInstance instance) {
     ApiDumpLayerRecordContent(contents);
 
     std::unique_lock<std::mutex> mlock(g_instance_dispatch_mutex);
-    XrGeneratedDispatchTable *next_dispatch = nullptr;
-    auto map_iter = g_instance_dispatch_map.find(instance);
-    if (map_iter != g_instance_dispatch_map.end()) {
-        next_dispatch = map_iter->second;
-    }
+    XrGeneratedDispatchTable *next_dispatch = g_instance_dispatch_map[instance];
```
rpavlik, line 564: Please revert this change (and any others like it) - it was recently included because of #53

rpavlik (changes requested):
In the files I made it through, I found one change. It's a bit hard to read with all the indentation changes - any way you could split this into multiple commits, where only the last one does the "simple" removal of try-catch?

Also, I'm hoping to land #69 soon which would collide with this, though probably making it somewhat simpler due to centralizing logic, just as a heads-up. Your review there would be appreciated.

zheqiMicrosoft:
I didn't realize that ApiDump is another layer. I will make a new PR with the same changes the ones in api_dump_generator.py and api_dump.cpp. And I will make sure indent change in a separate commit than the removal of try...catches. Thanks!

zheqiMicrosoft:
Hi Ryan @rpavlik, 
May I ask what's the estimate timeline of [#69](https://github.com/KhronosGroup/OpenXR-SDK/pull/69)? Thanks!

rpavlik:
Well, I got one approved review, hoping for another soon so I feel comfortable merging it. Also need to figure out the Windows build issue. If your revision of this lands first, I can rebase no problem, just wanted to mention it in case building on that one made the revision job any easier.

zheqiMicrosoft:
I see. Thanks!

rpavlik:
It's now merged. I can take a whack at porting your changes, if you like, otherwise if you make them on the old base because it works better for you, I can do the rebase for you and let you review it.

zheqiMicrosoft:
Thanks! I think I will just rebase it.

rpavlik:
I actually just looked at the diff in a viewer with whitespace ignored, and it is hopefully pretty easy to rebase - sorry about the mess I made for you.

Is this really all that the folks setting these requirements need - that try/catch and the manual throws are removed? Removing the manual throws makes sense, but the try/catch removal looks to me like removing the try/catch is risking an exception (thrown by a standard library component) escaping the ABI uncaught. That said, I've been lucky enough to usually be permitted exception usage, so I'm not very familiar with stringently avoiding them.

zheqiMicrosoft:
Yes. According to my understanding that’s all we need. https://github.com/KhronosGroup/OpenXR-SDK/issues/52 Maybe we can discuss on this thread so everyone can see it. Thanks for the help!

PR #72, opened 2019-06-26
filesystem: Fix theoretical bug on Linux.

rpavlik:
We'd try to close even a null directory.

Found by clang-tidy.

perhaps @ChristophHaag and/or @Ralith could review?


Ralith (approved):
LGTM, though perhaps it should return `false` as the try/catch logic seems to indicate is expected for errors? Use seems to be to skip walking the `vector` if false is returned, which is reasonable in that case.

rpavlik:
ok, that makes sense.

PR #73, opened 2019-06-26
Loader proper UNICODE support.

yl-msft:
Change to use "W" version of Win32 APIs for regkey, file path, etc. instead of "A" functions.
This is applied regardless of _UNICODE settings of the project, and always use UTF8 and UTF16 for string manipulations, which has better support in Win32 APIs for security, long file name, and globalization.

### DIFF
--- a/src/common/filesystem_utils.cpp
+++ b/src/common/filesystem_utils.cpp
@@ -20,6 +20,7 @@
 
 #include <cstring>
 #include "xr_dependencies.h"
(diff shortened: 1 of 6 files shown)


brycehutchings (commented):
Looked good to me, but someone else should look at it too

yl-msft:
Tested helloxr with this loader using latest OpenXR runtime for Mixed Reality.

rpavlik (approved):

PR #75, opened 2019-06-27
Clang tidy

rpavlik:
A variety of cleanups prompted by running clang-tidy (and include-what-you-use) over the code in src. Should be mostly simple/small things - no major refactoring this time.

### DIFF
--- a/src/api_layers/api_dump.cpp
+++ b/src/api_layers/api_dump.cpp
@@ -59,7 +59,7 @@ static ApiDumpRecordInfo g_record_info = {};
 static std::mutex g_record_mutex = {};
 
 // HTML utilities
-bool ApiDumpLayerWriteHtmlHeader(void) {
+bool ApiDumpLayerWriteHtmlHeader() {
     try {
         std::unique_lock<std::mutex> mlock(g_record_mutex);
         std::ofstream html_file;
@@ -168,7 +168,7 @@ bool ApiDumpLayerWriteHtmlHeader(void) {
     }
 }
 
-bool ApiDumpLayerWriteHtmlFooter(void) {
+bool ApiDumpLayerWriteHtmlFooter() {
     try {
         std::unique_lock<std::mutex> mlock(g_record_mutex);
         std::ofstream html_file;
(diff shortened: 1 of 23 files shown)


src/api_layers/api_dump.cpp
```diff
@@ -393,7 +399,7 @@ bool ApiDumpLayerRecordContent(std::vector<std::tuple<std::string, std::string,
     return success;
 }
 
-XrResult ApiDumpLayerXrCreateInstance(const XrInstanceCreateInfo *info, XrInstance *instance) {
+XrResult ApiDumpLayerXrCreateInstance(const XrInstanceCreateInfo * /*info*/, XrInstance * /*instance*/) {
```
daveh-lunarg, line 402: Actually surprised this even compiles.  But, if the params are unused, should just remove them - right?

src/api_layers/api_dump.cpp
```diff
@@ -589,7 +591,7 @@ extern "C" {
 
 // Function used to negotiate an interface betewen the loader and an API layer.  Each library exposing one or
 // more API layers needs to expose at least this function.
-LAYER_EXPORT XrResult xrNegotiateLoaderApiLayerInterface(const XrNegotiateLoaderInfo *loaderInfo, const char *apiLayerName,
+LAYER_EXPORT XrResult xrNegotiateLoaderApiLayerInterface(const XrNegotiateLoaderInfo *loaderInfo, const char * /*apiLayerName*/,
```
daveh-lunarg: Same comment, just remove unused ?

src/api_layers/api_dump.cpp
daveh-lunarg: Nevermind - this is shorthand for 'unused param', and signature cant change.  Learned something new!

daveh-lunarg (approved):
Not previously familiar with clang-tidy, but I do like the tidier version.

Except the brace locations - but that's google style clang-format's fault :(

src/api_layers/api_dump.cpp
rpavlik: yup yup :) It's like putting `_` at the beginning of a param/variable name in a variety of other languages (python, rust) - basically meaning "I know this isn't used, but I can't/won't change the signature so don't complain about it, compiler/linter."

rpavlik:
Clang tidy has definitely gotten more powerful since I had last applied its "fix" mode - what with entirely removing unused parameters for purely-internal (file-static) functions, really smart range-for conversion, etc. I was quite impressed - and quite bummed I had done some of those cleanups myself manually.

rpavlik:
I'll merge this after it finishes being reviewed in the internal GitLab.

rpavlik:
Incorporated in GitLab and will be pushed with next sync.

Issue #52, opened 2019-05-24
Loader should not use C++ exceptions


daveh-lunarg:
The existing loader code includes C++ exceptions as a part of it's error handling. OpenXR nominally follows the Google C++ style guide, which _weakly_ prohibits use of [exceptions](https://google.github.io/styleguide/cppguide.html#Exceptions).

In order to maintain widest possible compatibility, including with other projects which follow the Google style guide, the exception handling code should be removed from the loader. 

This public issue is to track resolution of the same topic in the OpenXR WG, [issue 974](https://gitlab.khronos.org/openxr/openxr/issues/974).

daveh-lunarg:
Assigned myself, but feel free to jump in anyone. I'm buried on another project for a couple weeks, will plan to work on this in July.

zheqiMicrosoft:
I can volunteer to make this change. From what we’ve been able to gather, there are three categories of exception throwing in the code.

1. Exceptions are thrown and caught directly by the function that throws.  These instances can be changed to early return with a corresponding XrResult and logging, if any. See src\api_layers\api_dump.cpp(473) in function ApiDumpLayerXrCreateApiLayerInstance.
2.  Throws that are the result of a catch(..) with some “unknown error” or “bad alloc” logging. These can be if defed out as the bad allocs likely come from the standard library and Chromium already has an OOM handler. 
3. Throw in core_validation.cpp.  This is only used by the validation layer that is not part of loader code.  We can ignore these instances for now.

All catches should be able to be if defed out.

@daveh-lunarg , do you agree with this assessment and next steps?

daveh-lunarg:
Yes, I agree with all 3, though I would go further and say don't #ifdef the exception code, just remove it completely. I'll add a new issue to track cleaning up the throw(s) in core_validation.cpp, so it doesn't get lost.

Also, prevent exceptions being thrown by the json library by clearing the `JSON_USE_EXCEPTION` definition in `external\jsoncpp\dist\json\json.h`.

rpavlik:
Instead of editing `external\jsoncpp\dist\json\json.h` I'd prefer setting `JSON_USE_EXCEPTION=0` in CMake: that way we could use system-provided jsoncpp or the vendored copy equivalently.

[...]

PR #76, opened 2019-06-28
Make Loader Exception Free

zheqiMicrosoft:
Try to make Loader exception free by doing the following:
1. For Exceptions are thrown and caught directly by the function that throws. These instances are changed to early return with a corresponding XrResult and logging, if any.
2. For Throws that are the result of a catch(..) with some “unknown error” or “bad alloc” logging. Remove these instances
3. Remove all catches directly.
4. Change JSON_USE_EXCEPTION to be defined as 0 so jsoncpp will not throw exception.

### DIFF
--- a/doc/loader/OpenXR_loader_design.html
+++ b/doc/loader/OpenXR_loader_design.html
@@ -3856,12 +3856,6 @@ std::experimental::filesystem::path search_path;</code></pre>
 </div>
 </div>
 <div class="paragraph">
-<div class="title">Exceptions</div>
-<p>The OpenXR loader relies on throwing and catching C++ exceptions.  Adding code
-should take this into account and properly wrap executing code with a "try"/"catch"
-block.</p>
-</div>
-<div class="paragraph">
 <div class="title">Experimental Filesystem Usage</div>
 <p>In order to simplify the file management, especially with regards to loading JSON
 manifest files or finding dynamic library files, the experimental/filesystem is used.
@@ -3885,13 +3879,10 @@ find elements of this functionality with the prefix
 <pre class="highlight"><code class="language-cpp" data-lang="cpp">#include &lt;experimental/filesystem&gt;
 
 static void checkAllFilesInThePath(const std::string &amp;search_path) {
-    try {
     // If the file exists, try to add it
     if (std::experimental::filesystem::is_regular_file(search_path)) {
         std::experimental::filesystem::path absolute_path =
             std::experimental::filesystem::absolute(search_path);
-    }
-    } catch (...) {
     }
 }</code></pre>
 </div>
@@ -5092,7 +5083,6 @@ same <code>LoaderInstance*</code>.</p>
     XrSession*                                  session) {
     bool primary_locked = false;
     bool secondary_locked = false;
-    try {
     g_instance_mutex.lock();
     secondary_locked = true;
     LoaderInstance *loader_instance = g_instance_map[instance];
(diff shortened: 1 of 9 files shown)


zheqiMicrosoft:
@rpavlik @daveh-lunarg Here is the new PR. The only difference with the previous one is that this one does not touch api_dump and things are separated into 2 commits. First commits adjust indentation, and second commit make the real change. Thanks!

rpavlik:
See issue #52

doc/loader/OpenXR_loader_design.html
```diff
@@ -3856,12 +3856,6 @@ <h4 id="_language_selection">6.3.2. Language Selection</h4>
 </div>
 </div>
 <div class="paragraph">
-<div class="title">Exceptions</div>
```
daveh-lunarg: This html is built from the loader design .adoc files in https://gitlab.khronos.org/openxr/openxr/tree/master/specification/loader. These changes should be made there or will be lost on next build/update.

daveh-lunarg (changes requested):
This all looks good to me.  I didn't test but by inspection I think the python changes do the right thing for the generated code as well.

Looks good to push pending removal of the html edits (and a corresponding MR for the loader .adoc files).

rpavlik:
@daveh-lunarg the loader adoc files are not public right now - this is why I wanted to make them public :)

zheqiMicrosoft:
Hi @daveh-lunarg , unfortunately I don't have access to the link you sent(https://gitlab.khronos.org/openxr/openxr/tree/master/specification/loader). Is there any other way I can modify the .adoc file? Thanks!

rpavlik:
@zheqiMicrosoft I'd just skip it - leave the modified HTML, and I'll open an issue internally to sync the source back. There's already a separate issue to publish the loader doc adoc files.

zheqiMicrosoft:
@rpavlik Hi Ryan, is there any other changes I need to make for this PR? Thanks!

zheqiMicrosoft:
@rpavlik @daveh-lunarg Hi Dave and Ryan, I just removed the doc/loader/OpenXR_loader_design.html file from this PR since it's generated. Thanks!

RafaelCintron:
Not having this change is a blocker for having an OpenXR backend in Chromium.  

@rpavlik and @daveh-lunarg how do we move forward on getting this in?

rpavlik:
I guess my main concern is that removing the catches isn't preventing exceptions, it's allowing exceptions to sneak out of the ostensibly-C ABI. The runtime shouldn't let exceptions escape, but there are standard library containers used and I haven't audited the calls (or added `nothrow`) to make sure that they won't throw.

RafaelCintron:
The version of the standard library that Chromium uses has been compiled to not throw exceptions.  This pull request eliminates throwing from the loader itself.  Is there any other exception throwing source we should be concerned about?  If not, I think we should move forward with this change.

rpavlik:
Ah, ok. Still makes me nervous we might leak exceptions to those who aren't using such a standard library, but I can probably propose a change later.

rpavlik:
OK, so I rebased it, then made some changes to allow non-Chromium users to avoid leaking exceptions - see #80. I'd rather wait for that whole thing (including this change) to merge than to merge this without the protection for non-chromium users, especially since chromium can presumably use a local fork right now anyway.

rpavlik:
Thanks for your work on this and your patience. I've rebased it into #80 which is now merged!

PR #80, opened 2019-07-24
Switchable exceptions

### LINKED ISSUE #52: Loader should not use C++ exceptions

rpavlik:
This accommodates the change from @zheqiMicrosoft and @rafaelcintron to remove try/catch from the loader. I added back only some function-level try/catch hidden in a macro, so that Chromium can build without it, but everybody else still can avoid leaking exceptions thru a C ABI.

Resolves #52


src/external/jsoncpp/include/json/config.h
```diff
@@ -22,7 +22,7 @@
 // If non-zero, the library uses exceptions to report bad input instead of C
 // assertion macros. The default is to use exceptions.
 #ifndef JSON_USE_EXCEPTION
-#define JSON_USE_EXCEPTION 1
```
thetuvix, line 25: Rather than changing the `JSON_USE_EXCEPTION` default to 0 by changing config.h in the JSON library, should we just set `JSON_USE_EXCEPTION` to 0 in the OpenXR code that includes the JSON library?

This change seems brittle for us to forget about and break later if we ever update the JsonCpp dependency.

rpavlik: Yeah I've wondered that too. I wasn't sure if it got compiled in (and thus needed to be in the build system) or if it was interface only (which seems unlikely). The advantage of this route is that if somebody uses their own build system, it will still be fine. Perhaps a preprocessor error in the loader code that checks this?

rpavlik: On the other hand, it is actually safe for jsoncpp to be built with exceptions given this change, since I added conditional catches around stuff - it's just not acceptable for the Chromium users.

rpavlik: OK, added an error that will fire if somebody is building in "chromium mode" but didn't disable exceptions in jsoncpp.

daveh-lunarg (approved):
This looks good. I'm good with the json warning, and like the ability to turn some exception catching on/off as desired. Having exceptions caught can hinder debugging sometimes, so the ability to toggle will be helpful.

Side note - OMG the github diff utility can make no f'ing sense of these changes.  Ugh.

src/loader/loader_logger.cpp
```diff
@@ -17,7 +17,6 @@
 // Author: Mark Young <marky@lunarg.com>
 //
 
-
```
daveh-lunarg: Whitespace change only - revert?

rpavlik:
yeah, I spent far too long de-confusing the git stuff. Mass indentation changes are apparently still a Hard Problem in 2019.

PR #84, opened 2019-07-29
api_dump: Fix Linux build.

rpavlik:
An earlier Windows fix accidentally repeatedly-defined this macro.


rpavlik:
No-op except on Linux, where I tested it.

PR #88, opened 2019-07-29
Remove unneeded generation

### LINKED ISSUE #74: Loader shouldn't provide API *ToString() implementations
Loader should not provide an implementation for ResultToString() and StructToString() API entry points. Either strip them out entirely, or retain for the loader's internal use but ensure they're not linked into the entry point's dispatch chain.

See https://gitlab.khronos.org/openxr/openxr/issues/1139

rpavlik:
The loader shouldn't be implementing the xr*ToString functions anymore, but it was. After that was removed, I could remove the xr_generated_utilities files.

Fixes #74 

Related to internal PR 1491 and internal issue 1139.


rpavlik:
Got internal review, merging here.

PR #81, opened 2019-07-25
Fix macro redefinition and include common_cmake_config.h errors

patrto:
- Build warning occurs when NOMINMAX or WIN32_LEAN_AND_MEAN is already defined
- If the loader files are generated without cmake, common_cmake_config.h doesn't exist so its inclusion should be wrapped around a macro.

PR #92, opened 2019-08-01
Fix Build Breaks

zheqiMicrosoft:
Build warning occurs when NOMINMAX or WIN32_LEAN_AND_MEAN is already defined
If the loader files are generated without cmake, common_cmake_config.h doesn't exist so its inclusion should be wrapped around a macro. google wants headers included alphabetically and loader_interfaces.h depends on openxr.h

### DIFF
--- a/src/common/loader_interfaces.h
+++ b/src/common/loader_interfaces.h
@@ -19,6 +19,8 @@
 
 #pragma once
 
+#include <openxr/openxr.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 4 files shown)


zheqiMicrosoft:
Duplicate of Patrto's PR(KhronosGroup/OpenXR-SDK-Source#81). So we can get it merged and unblocked.

rpavlik:
and it's legit that you're submitting this on his behalf? I don't want to get in trouble with lawyers on any side.

src/common/loader_interfaces.h
```diff
@@ -19,6 +19,8 @@
 
 #pragma once
 
+#include <openxr/openxr.h>
```
rpavlik, line 22: thanks for catching this. I had been slowly fixing these as I went, but absent a "every header must compile on its own" test it's not the easiest to fix en-masse.

PR #95, opened 2019-08-02
Guard config

rpavlik:
Consistently guard the config include - you have to define a symbol to cause the include, so that other-build-system-users get the right thing by default.

Look OK @zheqiMicrosoft @patrto? This should replace #81 and #92 but was independently done so I feel OK merging it.


zheqiMicrosoft:
LGTM. Thanks!

patrto:
LGTM, thank you Ryan!

PR #98, opened 2019-08-02
Export only core symbols

### LINKED ISSUE #45: Exported symbols from loader
When reviewing the loader docs, I decided to dump the list of exported symbols from the loader, when built as a dynamic library on Windows. More than I expected showed up - in particular, note the first few:

- `_LoaderGenTermXrResultToString@16`
- `_LoaderGenTermXrStructureTypeToString@16`
- `_LoaderXrTermCreateApiLayerInstance@12`
- `_LoaderXrTermCreateDebugUtilsMessengerEXT@16`
- `_LoaderXrTermCreateInstance@8`
- `_LoaderXrTermDestroyDebugUtilsMessengerEXT@8`
- `_LoaderXrTermDestroyInstance@8`
- `_LoaderXrTermGetInstanceProcAddr@16`
- `_LoaderXrTermSetDebugUtilsObjectNameEXT@12`
- `_LoaderXrTermSubmitDebugUtilsMessageEXT@28`
- `_xrAcquireSwapchainImage@16`
- `_xrApplyHapticFeedback@20`
- `_xrBeginFrame@12`
- `_xrBeginSession@12`
- `_xrConvertTimeToTimespecTimeKHR@20`
- `_xrConvertTimeToWin32PerformanceCounterKHR@20`
- `_xrConvertTimespecTimeToTimeKHR@16`
- `_xrConvertWin32PerformanceCounterToTimeKHR@16`
- `_xrCreateAction@16`
- `_xrCreateActionSet@16`
- `_xrCreateActionSpace@16`
- `_xrCreateDebugUtilsMessengerEXT@16`
- `_xrCreateInstance@8`
- `_xrCreateReferenceSpace@16`
- `_xrCreateSession@16`
- `_xrCreateSwapchain@16`
- `_xrDestroyAction@8`
- `_xrDestroyActionSet@8`
- `_xrDestroyDebugUtilsMessengerEXT@8`
- `_xrDestroyInstance@8`
- `_xrDestroySession@8`
- `_xrDestroySpace@8`
- `_xrDestroySwapchain@8`
- `_xrEndFrame@12`
- `_xrEndSession@8`
- `_xrEnumerateApiLayerProperties@12`
- `_xrEnumerateEnvironmentBlendModes@28`
- `_xrEnumerateInstanceExtensionProperties@16`
- `_xrEnumerateReferenceSpaces@20`
- `_xrEnumerateSwapchainFormats@20`
- `_xrEnumerateSwapchainImages@20`
- `_xrEnumerateViewConfigurationViews@32`
- `_xrEnumerateViewConfigurations@28`
- `_xrGetActionStateBoolean@20`
- `_xrGetActionStatePose@20`
- `_xrGetActionStateVector1f@20`
- `_xrGetActionStateVector2f@20`
- `_xrGetBoundSourcesForAction@20`
- `_xrGetCurrentInteractionProfile@20`
- `_xrGetD3D10GraphicsRequirementsKHR@20`
- `_xrGetD3D11GraphicsRequirementsKHR@20`
- `_xrGetD3D12GraphicsRequirementsKHR@20`
- `_xrGetInputSourceLocalizedName@36`
- `_xrGetInstanceProcAddr@16`
- `_xrGetInstanceProperties@12`
- `_xrGetOpenGLGraphicsRequirementsKHR@20`
- `_xrGetReferenceSpaceBoundsRect@16`
- `_xrGetSystem@16`
- `_xrGetSystemProperties@20`
- `_xrGetViewConfigurationProperties@24`
- `_xrGetVisibilityMaskKHR@24`
- `_xrGetVulkanDeviceExtensionsKHR@28`
- `_xrGetVulkanGraphicsDeviceKHR@24`
- `_xrGetVulkanGraphicsRequirementsKHR@20`
- `_xrGetVulkanInstanceExtensionsKHR@28`
- `_xrLocateSpace@28`
- `_xrLocateViews@28`
- `_xrPathToString@28`
- `_xrPerfSettingsSetPerformanceLevelEXT@16`
- `_xrPollEvent@12`
- `_xrReleaseSwapchainImage@12`
- `_xrResultToString@16`
- `_xrSessionBeginDebugUtilsLabelRegionEXT@12`
- `_xrSessionEndDebugUtilsLabelRegionEXT@8`
- `_xrSessionInsertDebugUtilsLabelEXT@12`
- `_xrSetDebugUtilsObjectNameEXT@12`
- `_xrSetInteractionProfileSuggestedBindings@12`
- `_xrStopHapticFeedback@16`
- `_xrStringToPath@16`
- `_xrStructureTypeToString@16`
- `_xrSubmitDebugUtilsMessageEXT@28`
- `_xrSyncActionData@16`
- `_xrThermalGetTemperatureTrendEXT@24`
- `_xrWaitFrame@16`
- `_xrWaitSwapchainImage@12`

cc @daveh-lunarg

rpavlik:
This adjusts the loader, on all platforms, to only export core functions, which is what I think we say in the docs we do. (aside from static linking, possibly only on windows, not sure)

This did require some changes to hello_xr, but I think only in accordance with the spec.

Not sure if we wanted to mark some of the graphics binding functions as exported too.

The other thing I did was "fix" the XRAPI_ATTR on Windows: it had gotten a declspec(dllexport) at some point, which I didn't find anywhere in Vulkan and I think was a mis-use of that attribute. We already had LOADER_EXPORT, it was just being inconsistently used and configured.

Have only tried this on Linux.

Fixes #45

### DIFF
--- a/include/openxr/openxr_platform_defines.h
+++ b/include/openxr/openxr_platform_defines.h
@@ -36,11 +36,7 @@ extern "C" {
  * Function pointer type: typedef void (XRAPI_PTR *PFN_xrFunction)(void);
  */
 #if defined(_WIN32)
-#ifdef XRAPI_DLL_EXPORT
-#define XRAPI_ATTR __declspec(dllexport)
-#else
 #define XRAPI_ATTR
-#endif
 // On Windows, functions use the stdcall convention
 #define XRAPI_CALL __stdcall
 #define XRAPI_PTR XRAPI_CALL
--- a/src/loader/CMakeLists.txt
+++ b/src/loader/CMakeLists.txt
@@ -37,6 +37,11 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     set(LOADER_NAME ${LOADER_NAME}-${MAJOR}_${MINOR})
 endif()
 
+if(NOT MSVC)
+    set(CMAKE_C_VISIBILITY_PRESET hidden)
+    set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+endif()
+
 # List of all files externally generated outside of the loader that the loader
 # needs to build with.
 SET(LOADER_EXTERNAL_GEN_FILES
@@ -168,6 +173,7 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
     target_compile_options(openxr_loader
         PRIVATE -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wpointer-arith
         PRIVATE -fno-strict-aliasing -fno-builtin-memcmp "$<$<COMPILE_LANGUAGE:CXX>:-fno-rtti>"
+        -ffunction-sections -fdata-sections
     )
     # For GCC version 7.1 or greater, we need to disable the implicit fallthrough warning since
     # there's no consistent way to satisfy all compilers until they all accept the C++17 standard
--- a/src/loader/loader_platform.hpp
+++ b/src/loader/loader_platform.hpp
@@ -31,6 +31,8 @@
 #define LOADER_EXPORT __attribute__((visibility("default")))
 #elif defined(__SUNPRO_C) && (__SUNPRO_C >= 0x590)
 #define LOADER_EXPORT __attribute__((visibility("default")))
+#elif defined(_WIN32) && defined(XRAPI_DLL_EXPORT)
+#define LOADER_EXPORT __declspec(dllexport)
 #else
 #define LOADER_EXPORT
 #endif
--- a/src/scripts/loader_source_generator.py
+++ b/src/scripts/loader_source_generator.py
@@ -126,6 +126,7 @@ class LoaderSourceOutputGenerator(AutomaticSourceOutputGenerator):
             preamble += '#include "openxr/openxr_platform.h"\n\n'
             preamble += '#include "loader_interfaces.h"\n\n'
             preamble += '#include "loader_instance.hpp"\n\n'
+            preamble += '#include "loader_platform.hpp"\n\n'
 
         elif self.genOpts.filename == 'xr_generated_loader.cpp':
             preamble += '#include "xr_generated_loader.hpp"\n\n'
@@ -213,6 +214,11 @@ class LoaderSourceOutputGenerator(AutomaticSourceOutputGenerator):
 
                     # Use the Cdecl directly from the XML
                     func_proto = cur_cmd.cdecl
+
+                    # Export only core functions
+                    if self.isCoreExtensionName(cur_cmd.ext_name):
+                        func_proto = func_proto.replace(
+                            "XRAPI_ATTR", "LOADER_EXPORT XRAPI_ATTR")
 
                     # Output the standard API form of the command
                     manual_funcs += func_proto
@@ -493,8 +499,14 @@ class LoaderSourceOutputGenerator(AutomaticSourceOutputGenerator):
 
                 if cur_cmd.protect_value:
                     generated_funcs += '#if %s\n' % cur_cmd.protect_string
+                decl = cur_cmd.cdecl.replace(";", " XRLOADER_ABI_TRY {\n")
 
-                generated_funcs += cur_cmd.cdecl.replace(";", " XRLOADER_ABI_TRY {\n")
+                # Export only core functions
+                if self.isCoreExtensionName(cur_cmd.ext_name):
+                    decl = decl.replace(
+                        "XRAPI_ATTR", "LOADER_EXPORT XRAPI_ATTR")
+
+                generated_funcs += decl
                 generated_funcs += tramp_variable_defines
 
                 # If this is not core, but an extension, check to make sure the extension is enabled.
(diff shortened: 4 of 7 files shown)


brycehutchings:
Why not export everything like it was? This sounds like a doc bug to me :-). The app is supposed to embed the loader with it, so exporting makes everything simpler. I could see why Vulkan didn't export extensions because each driver on the system might have different extensions and then you'd get linker errors, but that's avoided with OpenXR's. I really think we should keep exporting everything.

rpavlik:
That's a good point...

Do we maybe need to bring this to the group? Since there's at least some examples that already use get proc addr for their extension code, don't remember if there's anything in the actual spec itself that says this.

If we do go that route, we can just modify these changes since we still probably don't want to export essentially loader-internal symbols 😀

Tagging in @wallbraker and @ChristophHaag for Linux perspective

brycehutchings:
Agreed we should definitely not export internal stuff

brycehutchings:
I thought I had heard that the plan on Linux was to include the loader on the system rather than embed with the application. If so, that seems pretty concerning to me because once the various platforms diverge then decisions become a lot harder because there might not be a clear correct decision. I hope I misunderstood and Linux OpenXR apps include the loader like on the other platforms like Windows and Android.

rpavlik:
I thought that Android had a system wide loader too. It has to do with distributions - on Linux we do actually have a reasonable, centralized control so people don't step on each others libraries (distro package managers, etc). That, combined with the very rocky state of the loader on Linux early-on led us to switch the default to dynamic linking, to be able to fix breakages after shipping. This could be re-discussed, but that's probably a different issue than this one.

daveh-lunarg:
Not sure I follow - what are the *other* things to export, that aren't loader internal and aren't API? 

We can't export any extension symbols, so I can't think of what else there is?  Also, it appears to me that these changes to hello_xr are the correct way to find/call an extension entry point.

yl-msft:
"LOADER_EXPORT" seems exposing too many functions, see image below
Following function shouldn't be exported for openxr_loader.dll on windows platform, IMHO.
image

yl-msft (changes requested):
The key of exporting functions on win32 should be avoid exporting functions that's not OpenXR APIs.  Those currently decorated with XRAPI_ATTR are working fine today.  However, some internal loader function with "LOADER_EXPORT" decoration should be avoided being exported when compiling for Win32.

yl-msft:
BTW: I also agree with Bryce that we should exporting all XR functions in openxr headers, including all extensions that's in the spec.   Application shouldn't need to go through extra code to consume extensions if they are using the latest loader from this repo.

daveh-lunarg:
This is what I don't understand. Extensions are _always_ optional, and at loader compile time there's no way to know which extensions will be available at run time (regardless whether KHR, EXT, or vendor extension). I don't see how the loader can export any extension-defined symbol.

Agreed the `LoaderXr*` functions have to go.

yl-msft:
The loader code is generated from the same xr.xml that generated the corresponding openxr.h.   Therefore at loader compile time, it will support the same set of extensions in the openxr.h  They are always in sync.

It's by design loader exposes "optional" extensions.  If the runtime indicate any extension is not supported, it is invalid for app to call into those unsupported functions.

daveh-lunarg:
IMO it's risky to expose anything not guaranteed to exist on the platform.  Vulkan's loader exposes only core API and a constrained set of guaranteed-present extensions (e.g. WSI stuff in the Win DLL). Even that's a frequent problem, because people forget to enable the extension they need to use and the linker has no idea. Having entry points in the loader that work perfectly for developer A using runtime A, but later then crash with runtime B because he forget to check, is inviting pain.

It's a very crisp and logical division that you can call core APIs directly, but for any extension (which you already have to query for and enable) you also have to GetInstanceProcAddr for it's entry points. There's little to gain by skipping the GIPA for (some) extensions, and a lot of potential downside (IMO).

yl-msft:
OpenXR loader is designed differently than Vulkan.  The loader is not global on the system.  It is included with each application.    Therefore, the loader will not expose anything more than the corresponding openxr.h header already exposed, and vise versa.   As long as the app is compiled with openxr header, they are taken care of by the loader.

Also an OpenXR conformed runtime must not enable any extension unless it's explicitly enabled by the application in xrCreateInstance.  As long as the application is picking the extension name from the OpenXR.h, the loader must support this extension correctly.

Well if application is using an extension name outside of the OpenXR.h header, the loader cannot help them and they have to use GetInstanceProcAddr themselves and by passing loader logics.   But at least the behavior is consistent.

rpavlik:
Right now, in both branches (this and #100) the loader checks to make sure you've enabled an extension before passing the call to extension-created functions thru to the runtime. We've discussed removing this check from the loader, as it seems like "too much validation" in the loader vs. the validation layer, but it may be useful if we go with this more drastic change.

daveh-lunarg (approved):
Agreed I think 8/29/19 that core-only is correct, with an eye towards further restricting to just the 4 minimal-required set in near future.

yl-msft:
Agreed and approved.  
can i assume we will have a separate change to fix these?
image

yl-msft (approved):

rpavlik:
Yeah, presumably - I thought I had already gotten rid of those being exported. Is that in the static or dynamic library?

rpavlik:
I'm going to port this back to GitLab first to get our overall CI running on it, since this will probably mean we need some fixes to the sample code, etc.

daveh-lunarg:
This PR is approved and ready to push. I'm going to close issue #45 as resolved.

PR #106, opened 2019-08-02
List app changes

rpavlik:
cc @wallbraker - ensure I didn't mess anything up.

Just modernizes this little utility and installs it. Might want to change the name if we want to install it.


rpavlik:
Maybe rename to `openxr_info` ala `glinfo`

brycehutchings (approved):

PR #108, opened 2019-08-05
- Update json_stream initialization to improve compatibility

theblindprogramm3r:


src/loader/manifest_file.cpp
```diff
@@ -721,7 +721,7 @@ ApiLayerManifestFile::~ApiLayerManifestFile() = default;
 
 void ApiLayerManifestFile::CreateIfValid(ManifestFileType type, const std::string &filename,
                                          std::vector<std::unique_ptr<ApiLayerManifestFile>> &manifest_files) {
-    std::ifstream json_stream = std::ifstream(filename, std::ifstream::in);
+    std::ifstream json_stream(filename, std::ifstream::in);
```
rpavlik, line 724: Is the `std::ifstream::in` still required given that it's using a `std::ifstream`?

rpavlik (changes requested):
One minor question, plus we need the CLA

src/loader/manifest_file.cpp
theblindprogramm3r: Its not required, but by convention, it is recommended. Ex. if the default ever changes (not likely) from ::in to something else.

theblindprogramm3r: Ex. the default signature for ifstream:
explicit ifstream (const char* filename, ios_base::openmode mode = ios_base::in);

rpavlik: Interesting. I don't think I had used it with ifstream/ofstream in literally years. OK.

rpavlik (approved):

rpavlik:
Hmm, I clicked the recheck but it still says pending . Can you click the cla recheck link?

theblindprogramm3r:
> Hmm, I clicked the recheck but it still says pending . Can you click the cla recheck link?

I've gone back and done it, when I click on the link the top of the page says "You have signed the ...." and no longer offers a button to sign.

PR #111, opened 2019-08-14
Some fixes to Validation Layer (as found applying to the UE4 OpenXR plugin)

SteveSmithEpic:
Added a definition of LAYER_EXPORT for when building with Visual Studio so functions export correctly.
Fixed potential infinite loop in CoreValidationXrCreateSession if first binding is not a graphics binding
Added graphics binding types for Android, Win32, also Vulkan.

### DIFF
--- a/src/api_layers/core_validation.cpp
+++ b/src/api_layers/core_validation.cpp
@@ -46,6 +46,8 @@
 #define LAYER_EXPORT __attribute__((visibility("default")))
 #elif defined(__SUNPRO_C) && (__SUNPRO_C >= 0x590)
 #define LAYER_EXPORT __attribute__((visibility("default")))
+#elif defined(_MSC_VER)
+#define LAYER_EXPORT __declspec(dllexport)
 #else
 #define LAYER_EXPORT
 #endif
--- a/src/api_layers/core_validation.cpp
+++ b/src/api_layers/core_validation.cpp
@@ -593,9 +593,19 @@ XrResult CoreValidationXrCreateSession(XrInstance instance, const XrSessionCreat
         while (nullptr != cur_ptr) {
             switch (cur_ptr->type) {
                 default:
-                    continue;
+                    break;
+                case XR_TYPE_GRAPHICS_BINDING_VULKAN_KHR:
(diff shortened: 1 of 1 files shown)


src/api_layers/core_validation.cpp
```diff
@@ -591,9 +593,19 @@ XrResult CoreValidationXrCreateSession(XrInstance instance, const XrSessionCreat
         while (nullptr != cur_ptr) {
             switch (cur_ptr->type) {
                 default:
-                    continue;
```
daveh-lunarg, line 594: I wouldn't change this `continue` to `break`, as it will then fail to find any graphics binding that isn't first in the chain. We don't currently have any other types defined to chain to a SessionCreateInfo AFAIK, but it's not precluded by the spec.

daveh-lunarg (changes requested):
I don't like this fxn's logic much, which at first glance appears to be counting graphics bindings but in fact is just finding the first one in the chain. It should be simplified to 1) do nothing if headless is enabled or 2) verify at least one graphics binding is present.

That's not all required for this PR, but at min I think the default 'break' needs to be removed.

daveh-lunarg:
(Also, @SteveSmithEpic, we need you to click through the CLA link before we can take your PR)

src/api_layers/core_validation.cpp
SteveSmithEpic: The break here will break out of the switch, not the while (this was the initial bug I hit - I hit an infinite loop here on the continue since it wasn't checking for D3D11 so hit the continue and looped on the first item, never moving the iterator). By breaking here, it falls through to the "next" portion (line 628), and continues the iteration loop.

daveh-lunarg: Doh! That continue sucked me in.  You're right, needs a break.

daveh-lunarg (approved):

src/api_layers/core_validation.cpp
SteveSmithEpic: It got me too - it's a wily one. ;)

daveh-lunarg:
Merged. Thanks!

PR #112, opened 2019-08-17
Update active runtime search documentation 

Ralith:
Summarizes [the logic in manifest_file.cpp](https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/master/src/loader/manifest_file.cpp#L315-L359) as of #51.

CC @ChristophHaag

### DIFF
--- a/specification/Makefile
+++ b/specification/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


rpavlik:
Looks good to me - out of curiosity, what environment did that makefile not work in?

rpavlik (approved):

PR #114, opened 2019-08-20
Support for building WindowsStore loader and layers and simplified filename

brycehutchings:
CMake requires some extra parameters when building for WindowsStore (UWP), which is used by HoloLens (and limited use on Desktop). To build targeting UWP, "-DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION=10.0" is passed to CMake. 

From experimentation, I found "WIN32" is true for both Windows and WindowsStore, so I used it where the two could be treated the same.

I also added a new option BUILD_ALL_EXTENSIONS which will cause CMake to fail if any extension isn't enabled that could be enabled for the platform (AFAIK, this just means Vulkan and OpenGL support). I did this because we want to have a way to ensure binaries can be built that can be distributed for all to use without accidentally removing support (e.g. building on a machine that doesn't have the Vulkan SDK).

Finally, the loader has a special file name on Windows and not on other platforms (openxr_loader-1_0 versus openxr_loader on all other platforms). I am unsure why this is, but given that Windows apps bundle the loader, I see no reason to make the filename versioned. It will just potentially make it harder to upgrade depending on how apps do their packaging and linking.


src/loader/CMakeLists.txt
```diff
@@ -130,24 +129,27 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
...
-                             CMAKE_C_FLAGS_RELWITHDEBINFO
-                             CMAKE_CXX_FLAGS_DEBUG
-                             CMAKE_CXX_FLAGS_RELEASE
-                             CMAKE_CXX_FLAGS_RELWITHDEBINFO)
-        # If building DLLs, force static CRT linkage
-        if(DYNAMIC_LOADER)
-            if (${configuration} MATCHES "/MD")
-                string(REGEX REPLACE "/MD" "/MT" ${configuration} "${${configuration}}")
+elseif(WIN32)
+    # WindowsStore (UWP) apps must be compiled with dynamic CRT linkage (default)
```
yl-msft, line 133: Not: Is it better to verify this default?

yl-msft (approved):

src/loader/CMakeLists.txt
brycehutchings: I'd rather do less until there is a need to do more. I don't want to make the build scripts more complicated than they need to be.

rpavlik:
Would it make sense to also take care of letting the CRT linkage type be configurable in this commit?

I can go back in GitLab and see when/who added the version to the filename.

rpavlik (approved):
Looks good to me. The only reason I can think of to leave the version in there, is that it might be useful to see the version in the filename - Linux has sonames that automatically get appended.

PR #116, opened 2019-08-22
More cleanup

rpavlik:
Additional static analysis fixes and other minor cleanup. Builds on #115


brycehutchings (approved):

rpavlik:
the capitalization was from this: https://clang.llvm.org/extra/clang-tidy/checks/readability-uppercase-literal-suffix.html  motivation being that `8l` looks a lot like `81` despite being `8L`, presumably. Not sure why it went overboard, but I fixed it.

PR #118, opened 2019-08-23
Fix logic error in Linux active runtime search

Ralith:
When XDG_CONFIG_DIRS was unset, the fallback directories were ignored, resulting in /etc/xdg not being searched by default.

In retrospect I'm not sure what I was thinking with the original GetXDGEnv, that was way too complicated and it bit us.

### DIFF
--- a/src/loader/manifest_file.cpp
+++ b/src/loader/manifest_file.cpp
@@ -256,32 +256,34 @@ static void ReadDataFilesInSearchPaths(ManifestFileType type, const std::string 
 
 #ifdef XR_OS_LINUX
 
-// If ${name} has a nonempty value, return it; if both other arguments are supplied return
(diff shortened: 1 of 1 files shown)


rpavlik (approved):
Looks good!

rpavlik:
So this in practice results in the following paths on my system being searched:

- `/home/ryan/.config/openxr/1/active_runtime.json`
- `/etc/xdg/openxr/1/active_runtime.json`
- `/usr/local/etc/openxr/1/active_runtime.json`
- `/etc/openxr/1/active_runtime.json`

which seems reasonable enough - the etc/xdg dir is newer standard so it comes ahead of the older stuff, and /usr/local/etc (or whatever your install prefix is) precedes /etc

Thanks for troubleshooting this with me.

PR #119, opened 2019-08-23
Loader test update

daveh-lunarg:
The loader_test is a unit test of loader functionality, including debug messenger. It was previously only run against the faux-runtime binaries and only on Linux, and had fallen into disrepair. This switches to using the installed runtime, if one is available, and otherwise degrades gracefully to a small number of loader-handled checks only.

This version has only been tested to date on Win10 and on a Linux system with no active runtime (skipping most tests). The faux-runtime allowed _null_ devices, but the active runtime tests required creation of an `ID3D11Device` during test. Corresponding device creation code will most likely be required in the Linux paths.
 
Intent is to use it as an automated CI check, once complete.

The bulk of the changes are in loader_test.cpp, but some of the refactoring and bug-fixes ended up touching a relatively large number of files.


daveh-lunarg:
Fixed the build issue, but now have a couple test failures on Win10.  Will review on Mon.

src/api_layers/CMakeLists.txt
```diff
@@ -97,10 +102,15 @@ endif()
...
+if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
+    set(LAYER_LIB_NAME "XrApiLayer_core_validation")
+else()
+    set(LAYER_LIB_NAME "libXrApiLayer_core_validation")
+endif()
 gen_xr_layer_json(
     ${CMAKE_CURRENT_BINARY_DIR}/XrApiLayer_core_validation.json
     LUNARG_core_validation
-    libXrApiLayer_core_validation${CMAKE_SHARED_MODULE_SUFFIX}
+    ${LAYER_LIB_NAME}${CMAKE_SHARED_MODULE_SUFFIX}
```
rpavlik: I wonder if a CMake "generator expression" would work here... `$<TARGET_LINKER_FILE_NAME:XrApiLayer_core_validation>` instead of `${LAYER_LIB_NAME}${CMAKE_SHARED_MODULE_SUFFIX}`

src/common/object_info.cpp
```diff
@@ -238,44 +241,45 @@ NamesAndLabels DebugUtilsData::PopulateNamesAndLabels(std::vector<XrSdkLogObject
     return {objects, labels};
 }
 
-AugmentedCallbackData DebugUtilsData::AugmentCallbackData(
-    const XrDebugUtilsMessengerCallbackDataEXT& provided_callback_data) const {
-    AugmentedCallbackData ret{provided_callback_data};
-    if (object_info_.Empty() || provided_callback_data.objectCount == 0) {
-        return ret;
+void DebugUtilsData::WrapCallbackData(AugmentedCallbackData* aug_data,
```
rpavlik, line 244: why all the pointers? Are some of these expected/permitted to be null? If not, switch to references, safer and clearer indication of intent. Also, not sure why you switched to an out-param instead of a return value.

src/common/object_info.cpp
```diff
@@ -238,44 +241,45 @@ NamesAndLabels DebugUtilsData::PopulateNamesAndLabels(std::vector<XrSdkLogObject
...
-        return ret;
+    // If we found nothing to add, return the original data
+    if (!name_found && aug_data->labels.empty()) {
+        return;
     }
 
-    // If a name or a label has been found, we should update it in a new version of the callback data
-    ret.new_objects.assign(provided_callback_data.objects, provided_callback_data.objects + provided_callback_data.objectCount);
+    // Found additional data - modify an internal copy and return that as the exported data
+    memcpy(&aug_data->modified_data, callback_data, sizeof(XrDebugUtilsMessengerCallbackDataEXT));
```
rpavlik, line 271: why did you switch from the C++ assign to memcpy?

src/loader/api_layer_interface.cpp
```diff
@@ -64,13 +64,24 @@ XrResult ApiLayerInterface::GetApiLayerProperties(const std::string& openxr_comm
...
+                                              "VUID-XrApiLayerProperties-type-type: unknown type in api_layer_properties");
+                return XR_ERROR_VALIDATION_FAILURE;
+            }
+        }
+    }
+
     // Find any implicit layers which we may need to report information for.
     XrResult result = ApiLayerManifestFile::FindManifestFiles(MANIFEST_TYPE_IMPLICIT_API_LAYER, manifest_files);
-    if (XR_SUCCESS == result) {
+    if (XR_SUCCEEDED(result)) {
```
rpavlik, line 80: Is there a non-XR_SUCCESS return code that's expected here? (Since this is a semantics change - it's now `result >= 0` )

src/loader/loader_core.cpp
```diff
@@ -251,16 +252,50 @@ LOADER_EXPORT XRAPI_ATTR XrResult XRAPI_CALL xrCreateInstance(const XrInstanceCr
         return result;
     }
 
+    // Remove any loader-provided extensions (probably just debug_utils) from the instance_create_info
```
rpavlik: so I think the logic here is wrong, since the app might have enabled debug utils and want the runtime to implement it. Should strip it only if the runtime doesn't implement it. In any case, this change is big enough it should be reviewed separately.

src/loader/loader_core.cpp
```diff
@@ -251,16 +252,50 @@ LOADER_EXPORT XRAPI_ATTR XrResult XRAPI_CALL xrCreateInstance(const XrInstanceCr
...
+                break;
+            }
+        }
+        if (!is_loader_ext) {
+            runtime_ext_names.push_back(info->enabledExtensionNames[i]);
+        }
+    }
+
+    // Build a temporary instance create_info containing the purged list of extensions
+    const char* purged_ext_list[1024];  // Just make it absurdly large
```
rpavlik: not really necessary, since apparently the conditions for invalidating the `.c_str()` of a std::string are narrower than I had thought til Paul set me straight :D  https://en.cppreference.com/w/cpp/string/basic_string/c_str

> The pointer obtained from c_str() may be invalidated by:
>
> -   Passing a non-const reference to the string to any standard library function, or
> -   Calling non-const member functions on the string, excluding operator[], at(), front(), back(), begin(), rbegin(), end() and rend(). 

So, just use a std::string.

src/loader/loader_platform.hpp
```diff
@@ -150,8 +150,31 @@ static LoaderPlatformLibraryHandle LoaderPlatformLibraryOpen(const std::string &
...
-            // If that failed, then try loading it with broader search folders.
+            // DLL file was found, but it didn't load. Try some broader search options in case
+            // there was a dependency that wasn't found.
             handle = LoadLibraryExW(pathW.c_str(), NULL, LOAD_LIBRARY_SEARCH_DEFAULT_DIRS | LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR);
+        } else {
+            // DLL file was not found. In case it's a relative path, add the override envars to the
+            // search path and try again. Add both runtime and API layer paths, if set
+            DLL_DIRECTORY_COOKIE rt_cookie = 0, api_cookie = 0;
+
+            std::string envar_path = PlatformUtilsGetSecureEnv(OPENXR_RUNTIME_JSON_ENV_VAR);
```
rpavlik: If you could move this change to internal GitLab, I'd like the MS folks to have a look at it from a security PoV.

src/scripts/generate_runtime_manifest.py
```diff
@@ -45,15 +45,17 @@ def main(argv):
...
-    file_text  = '{\n'
-    file_text += '    "file_format_version": "%s",\n' % cur_runtime_json_version
-    file_text += '    "runtime": {\n'
-    file_text += '        "library_path": "%s"\n' % library_location
-    file_text += '    }\n'
-    file_text += '}\n'
-    f = open(output_file, 'w')
-    f.write(file_text)
-    f.close()
+	# Don't need a fake 'good' manifest - use the active (installed) runtime
```
rpavlik: I'd rather just remove this code if it's not needed anymore - it's in Git if we want it back.

rpavlik (commented):
I need to actually check out these changes, rather than review a diff, but I've got some comments in-line. Thanks for this work - there's a lot of changes here. There are some that would be great to get submitted and reviewed separately.

Are all the `XR_SUCCESS == result` -> `XR_SUCCEEDED(result)` just changed for convention's sake, or are there actual success-but-not-XR_SUCCESS result codes you were hitting?

src/common/object_info.cpp
daveh-lunarg: This was a memory corruption bug that was very hard to track down.  Previous behavior was returning a local stack variable that was out of scope to the caller, which then worked for an indeterminate amount of time, until the un-allocated memory eventually got overwritten. 

This change requires the calling scope to allocate the output structure.

src/common/object_info.cpp
daveh-lunarg: I basically re-wrote the entire function, and memcpy() is clearer to me. No significant preference, only familiarity.

src/loader/api_layer_interface.cpp
daveh-lunarg: Result of a global search on `XR_SUCCESS ==`, I think.  The SUCCEEDED macro is no less correct, and more future-proof (IMO).

src/loader/loader_core.cpp
daveh-lunarg: Yeah, I think you're right.  We don't want to return `UNIMPLEMENTED` if the loader supports the extension, regardless of whether the runtime does. But we also _do_ want to enable it on the runtime if it _is_ supported, which this change prevents.

I'll back this out, and probably put it up as a separate PR.

src/scripts/generate_runtime_manifest.py
daveh-lunarg: Agreed

src/api_layers/CMakeLists.txt
daveh-lunarg: That's better - done.

src/loader/loader_core.cpp
daveh-lunarg: Don't follow exactly?  But this code is moving to a separate PR.

src/loader/loader_platform.hpp
daveh-lunarg: Agreed

daveh-lunarg:
I've moved two areas of discussion out into a separate PR #123, and gitlab MR #[1540](https://gitlab.khronos.org/openxr/openxr/merge_requests/1540). 

The remaining code here updates the loader test to be more functional, but isn't completely so until the above PR/MRs are included. Pushing this now as a "no harm" PR.

rpavlik:
fyi, if you can, please run `./runClangFormat.sh` before committing. Makes my life easier when it comes time to sync back to GitLab :)

daveh-lunarg:
Yeah, and sorry.  I use the Visual Studio clang-format built-in, but must have slipped my mind here.

PR #123, opened 2019-08-29
(WIP) Allow loader-supplied extension to succeed in CreateInstance

daveh-lunarg:
The loader provides the bulk of the debug messenger functionality internally, but the runtime may also either 1) participate, 2) provide it's own debug messenger, or 3) neither.  In all 3 cases we'd like the CreateInstance to succeed when DbgMsg is specified.

This PR in current state just removes DbdMsg from the extension list before passing it on to the runtime, which fixes case 3 but disallows 1 & 2.


rpavlik:
#129 merged so this can be closed.

PR #125, opened 2019-08-30
Include a debug OutputDebugString logger for Win32

brycehutchings:
The loader automatically enables a stderr logger, but this isn't visible in non-Console win32 apps (WinMain). This adds a default debugger output for _WIN32 (OutputDebugString) which should be more useful.

### DIFF
--- a/src/loader/loader_logger_recorders.cpp
+++ b/src/loader/loader_logger_recorders.cpp
@@ -28,9 +28,52 @@
 #include <string>
 #include <vector>
 #include <iostream>
+#include <sstream>
+
+#ifdef _WIN32
+#include <Windows.h>
+#endif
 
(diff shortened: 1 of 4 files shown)


src/loader/loader_logger_recorders.cpp
```diff
@@ -28,9 +28,52 @@
 #include <string>
 #include <vector>
 #include <iostream>
+#include <sstream>
+
+#ifdef _WIN32
```
rpavlik, line 33: I can't keep track of which ifdef's we're using - this looks reasonable enough to me but I have no idea if it matches the other code offhand.

rpavlik (approved):
Hey, this looks pretty great! Hopefully you found it easy enough to add after I did the refactoring of this code recently.

PR #127, opened 2019-09-03
validation layer: MND headless

rpavlik:
Fix validation layer in the presence of the Monado headless extension


rpavlik:
Oops, this is already in #116

PR #128, opened 2019-09-03
loader-test: Clang-tidy, with some manual adjustment afterward.

rpavlik:


daveh-lunarg (approved):
These all look good to me

PR #129, opened 2019-09-04
Remove debug ext only when added by loader

### LINKED ISSUE #124: xrCreateInstance failed when requesting XR_EXT_debug_util extension
the loader added XR_EXT_debug_util extension to the list of runtime supported extensions when app is calling xrEnumerateSupportedExtensions function.

However, when app call xrCreateInstance, the loader didn't remove "XR_EXT_debug_util" extension from the creation info, and runtime get this extension in the list.  If the runtime didn't implement this extension, (which it typically won't because loader already did this), then xrCreateInstance will fail.

Expectation is when a layer (in this case the loader) inserted an additional extension to the extension support list, it should remove this item in the xrCreateInstance call so the lower layer get what it actually supports.

rpavlik:
Fixes #124 without the side-effects of the other approach (that is, this is an alternate to #123)

cc @yl-msft 

This can be tested with Monado - the master branch reports the support for the extension (though it's not fully implemented), while https://gitlab.freedesktop.org/monado/monado/commits/no-debug-messenger has it disabled for testing purposes.

To test, you can modify hello_xr to require the debug extension as follows:

Find this code in `openxr_program.cpp`:

```c++

        // Create union of extensions required by platform and graphics plugins.
        std::vector<const char*> extensions;
```

and change it to this:

```c++
        // Create union of extensions required by platform and graphics plugins.
        std::vector<const char*> extensions = {XR_EXT_DEBUG_UTILS_EXTENSION_NAME};
```

Note that the `InstanceCreateInfoManager` is intended to be re-usable in other layers, etc that may have a need for pruning the extension list. Also, the current state in this branch is that the extension is removed from the create info even if there are layers that report support for it (none of the built-in ones do so despite actually working with it?) - this might need some fine-tuning. However, it will avoid spurious failures.

### DIFF
--- a/src/loader/loader_instance.cpp
+++ b/src/loader/loader_instance.cpp
@@ -69,18 +69,14 @@ class InstanceCreateInfoManager {
     // Remove extensions named in the parameter and return a pointer to the current state.
     const XrInstanceCreateInfo* FilterOutExtensions(const std::vector<const char*>& extensions_to_skip) {
         if (enabled_extensions_cstr.empty()) {
-            return &modified_create_info;
+            return Get();
         }
         if (extensions_to_skip.empty()) {
-            return &modified_create_info;
+            return Get();
         }
-        auto b = enabled_extensions_cstr.begin();
-        auto e = enabled_extensions_cstr.end();
-        auto shouldSkipExtension = [&](const char* extensionInQuestion) {
-            auto it = std::find_if(b, e, [&](const char* extension) { return strcmp(extensionInQuestion, extension) == 0; });
-            return it != extensions_to_skip.end();
-        };
-        vector_remove_if_and_erase(enabled_extensions_cstr, shouldSkipExtension);
+        for (auto& ext : extensions_to_skip) {
+            FilterOutExtension(ext);
+        }
         return Update();
     }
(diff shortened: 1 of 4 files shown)


src/loader/loader_instance.cpp
```diff
@@ -48,6 +48,68 @@ const std::array<XrExtensionProperties, 1>& LoaderInstance::LoaderSpecificExtens
...
+        if (extensions_to_skip.empty()) {
+            return Get();
+        }
+        for (auto& ext : extensions_to_skip) {
+            FilterOutExtension(ext);
+        }
+        return Update();
+    }
+    // Remove the extension named in the parameter and return a pointer to the current state.
+    const XrInstanceCreateInfo* FilterOutExtension(const char* extension_to_skip) {
```
yl-msft, line 83: Nit: it seems the Get() and FilterOutExtension (w/o 's') functions can be private.

yl-msft (approved):

src/loader/loader_instance.cpp
rpavlik: Those are intended for use in "re-use" of this class elsewhere. e.g. if there's a layer providing an overlay extension, it would want to filter out just that extension before passing things down the line.

PR #132, opened 2019-09-13
Fix directory searching for layers

brycehutchings:
The usage of `FindFirstFileW` was incorrectly assuming it just takes a path to a directory. It also requires a filename (which can be a wildcard). In the process of fixing FileSysUtilsFindFilesInPath I found that the std::string that was passed in had a length different than where the null terminator was, which made appending a string break. This null terminator bug was traced to the code that reads the environment variable.


brycehutchings:
I apparently did this fix with old code and it has already been partially fixed (although using FileSysUtilsCombinePaths is a bit cleaner). The null-terminator fix is also needed AFAIK.

rpavlik (approved):

PR #133, opened 2019-09-13
Remove shwapi dependency

brycehutchings:
Rather than use the shwapi APIs (PathIsDirectoryW/PathFileExistsW) this changes to use GetFileAttributesW instead (kernel32.dll). This removes a dependency and also makes the loader WACK (Windows App Cert Kit) compliant.

### DIFF
--- a/src/common/filesystem_utils.cpp
+++ b/src/common/filesystem_utils.cpp
@@ -152,7 +152,10 @@ bool FileSysUtilsFindFilesInPath(const std::string& path, std::vector<std::strin
 
 #elif defined(XR_OS_WINDOWS)
 
-bool FileSysUtilsIsRegularFile(const std::string& path) { return !FileSysUtilsIsDirectory(path); }
+bool FileSysUtilsIsRegularFile(const std::string& path) { 
+    const DWORD attr = GetFileAttributesW(utf8_to_wide(path).c_str());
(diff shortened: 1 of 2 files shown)


src/common/filesystem_utils.cpp
```diff
@@ -152,14 +152,16 @@ bool FileSysUtilsFindFilesInPath(const std::string& path, std::vector<std::strin
 
 #elif defined(XR_OS_WINDOWS)
 
-// For pre C++17 compiler that doesn't support experimental filesystem
```
yl-msft, line 155: This line of comment can remain.  It is to comment about fallback from std::filesystem.

src/common/filesystem_utils.cpp
```diff
@@ -152,14 +152,16 @@ bool FileSysUtilsFindFilesInPath(const std::string& path, std::vector<std::strin
...
-#include <shlwapi.h>
+bool FileSysUtilsIsRegularFile(const std::string& path) { return !FileSysUtilsIsDirectory(path); }
```
yl-msft: if "is_dir" = (path exist) && (attr has dir), then !(is_dir) = !(path exist) || !(attr has dir).

"is regular file" here might be incorrectly true for non exist path, even if the path is not a file.

yl-msft (approved):

rpavlik (approved):
Looks reasonable to me, though I didn't test it. I suspect that trusting an MS dev to know Windows APIs is safe :)

Issue #136, opened 2019-09-23
Validation layer for xrEnumerateViewConfigurations incorrectly applies validation to the output buffer


jherico:
`xrEnumerateViewConfigurations` is a two-call-idiom function.  A simplified example of using it is....

```c++
    uint32_t viewConfigurationTypeCountOutput = 0;
    uint32_t viewConfigurationTypeCapacityInput = 0;
    xrEnumerateViewConfigurations(instance, systemId,  viewConfigurationTypeCapacityInput, &viewConfigurationTypeCountOutput, nullptr));
    std::vector<XrViewConfigurationType> output;
    output.resize(viewConfigurationTypeCountOutput);
    viewConfigurationTypeCapacityInput = viewConfigurationTypeCountOutput;
    result = xrEnumerateViewConfigurations(instance, systemId,  viewConfigurationTypeCapacityInput, &viewConfigurationTypeCountOutput, output.data()));
```

However, this code is triggering a validation error, because the validation logic is trying to verify that the contents of the output buffer are valid enums _before_  the call to the implementation.  However, calling `output.resize()` here default initializes the new values to 0, which is not one of the valid `XrViewConfigurationType` enums.   

While it may be reasonable to check the structure type and chain information for output buffers that return extensible structures, it's not reasonable to check other members, or in the case of enums like this to check the enum value for an output buffer.  That data is specifically intended to be overwritten.

rpavlik:
closed by #138, not sure how I missed closing it.

PR #138, opened 2019-09-23
Don't validate output enum buffer values

jherico:
Fix for #136


rpavlik:
Looks good on inspection, will test with an app.

rpavlik:
Reproduced #136 and confirmed this fixes it.

PR #139, opened 2019-09-23
Write output atomically at the end

jherico:
The behavior of the generator, to truncate it's output file at the start and then to write it all at the end is annoying when you're debugging one of the generators and also have the output file open in an editor in the debugger, because it keeps forcing the open editor back to the start of an empty file.

This PR resolves the problem by writing output to a tempfile and then at the end atomically replacing existing files with the tempfile.  This allows an open editor in tools like VS Code to retain their existing position in the file.

### DIFF
--- a/specification/scripts/generator.py
+++ b/specification/scripts/generator.py
@@ -580,7 +580,7 @@ class OutputGenerator:
                 directory = Path(self.genOpts.directory)
                 if not Path.exists(directory):
                     os.makedirs(directory)
-            shutil.copy(self.outFile.name, self.genOpts.directory + '/' + self.genOpts.filename)
+            shutil.move(self.outFile.name, self.genOpts.directory + '/' + self.genOpts.filename)
         self.genOpts = None
 
     def beginFeature(self, interface, emit):
--- a/specification/scripts/generator.py
+++ b/specification/scripts/generator.py
@@ -21,7 +21,9 @@ import io
 import os
 import pdb
 import re
+import shutil
 import sys
+import tempfile
 try:
     from pathlib import Path
 except ImportError:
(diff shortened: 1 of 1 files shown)


rpavlik:
cc @oddhack for shared tooling.

FWIW, it seems to work well for the openxr-hpp stuff.

oddhack:
Wouldn't shutil.move be better than .copy? Principle seems OK.

jherico:
My reading of `shutil.move` is that if the destintation doesn't exist it will be treated as a directory path, and the existing file moved to that newly created directory with the existing name, so that's not suitable.

I may be reading incorrectly.

oddhack:
I'm not stuck on the specific mechanism, it just seems undesirable to leave temporary files in place. Unless I misunderstand something about the copy() method.

oddhack:
Also, I just tried shutil.move('foo','bar') in python3 and the existing file 'foo' was correctly moved to the nonexistent file 'bar' in the same directory, which was what I had hoped for.

specification/scripts/generator.py
```diff
@@ -580,6 +575,12 @@ def endFile(self):
         self.outFile.flush()
         if self.outFile != sys.stdout and self.outFile != sys.stderr:
             self.outFile.close()
+        if self.genOpts.filename is not None:
+            if sys.platform == 'win32':
+                directory = Path(self.genOpts.directory)
+                if not Path.exists(directory):
+                    os.makedirs(directory)
+            shutil.copy(self.outFile.name, self.genOpts.directory + '/' + self.genOpts.filename)
```
jherico:
```suggestion
            shutil.move(self.outFile.name, self.genOpts.directory + '/' + self.genOpts.filename)
```

jherico:
changed to move

specification/scripts/generator.py
```diff
@@ -580,6 +575,12 @@ def endFile(self):
         self.outFile.flush()
         if self.outFile != sys.stdout and self.outFile != sys.stderr:
             self.outFile.close()
+        if self.genOpts.filename is not None:
+            if sys.platform == 'win32':
```
rpavlik, line 579: what's the reason for this platform conditional?

jherico: I don't know exactly, I just wanted to reproduce the behavior from the old code.

rpavlik: oh, ok. sounds good, thanks.

PR #141, opened 2019-10-23
Support system libs better

rpavlik:
This makes it possible to build against system jsoncpp. Required for some packaging.

It also makes it possible to drop the wglext.h and glext.h headers from the source package if, for instance, you're on a platform that has those headers system-installed and you don't want to duplicate things...

With these patches, if your distribution requires it, you may drop the following files/directories from the source tree and, as long as you have the associated deps system-installed, still succeed in building:

- `external/include/GL/wglext.h`
- `external/include/GL/glext.h`
- `external/python`
- `external/libs` (I suspect this is entirely unused...)
- `src/external/jsoncpp`
- `specification/scripts/katex` (I think not needed anywhere)
- `specification/scripts/htmldiff` (only needed on docs repos, I think)

### DIFF
--- a/src/loader/CMakeLists.txt
+++ b/src/loader/CMakeLists.txt
@@ -80,12 +80,21 @@ add_library(openxr_loader ${LIBRARY_TYPE}
     ${PROJECT_SOURCE_DIR}/src/common/object_info.cpp
     ${PROJECT_SOURCE_DIR}/src/common/object_info.h
     ${PROJECT_SOURCE_DIR}/src/common/platform_utils.hpp
-    ${PROJECT_SOURCE_DIR}/src/external/jsoncpp/src/lib_json/json_reader.cpp
-    ${PROJECT_SOURCE_DIR}/src/external/jsoncpp/src/lib_json/json_value.cpp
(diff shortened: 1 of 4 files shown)


src/loader/CMakeLists.txt
```diff
@@ -140,7 +145,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     endif()
 
     set_target_properties(openxr_loader PROPERTIES SOVERSION "${MAJOR}" VERSION "${MAJOR}.${MINOR}.${PATCH}")
-    target_link_libraries(openxr_loader -lstdc++fs -ldl -lpthread -lm)
+    target_link_libraries(openxr_loader PRIVATE stdc++fs PUBLIC dl pthread m)
```
bradgrantham-lunarg, line 149: Why private for filesystem and public for dl, pthread, and math for this PR?  (Could be fine, just asking)

rpavlik: Because if you don't have the consuming app link against pthread (and probably dl), it will fail to link. However, the fs usage should be entirely internal.

It's mostly moot right now since we aren't exporting a cmake config file from this, so it only applies to stuff in this package, but essentially public libs are the "interface" libs that would show up in e.g. a pkgconfig file, etc.

PR #143, opened 2019-11-04
Define XR_USE_GRAPHICS_API_OPENGL_ES

emersion:
This allows the loader to expose `XR_KHR_opengl_es_enable`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -99,6 +99,13 @@ elseif(BUILD_ALL_EXTENSIONS)
     message(FATAL_ERROR "OpenGL not found")
 endif()
 
+find_package(OpenGLES)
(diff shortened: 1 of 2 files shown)


rpavlik:
Hmm - it's a design flaw that it even needs this, and we'll presumably fix that soon. The find module needs some care (and a license header) so I'll defer this for now.

rpavlik:
Can you check and see if this is still needed with #146?

rpavlik:
OK, I think this is no longer needed with the loader in 1.0.6, but lack the system to check. Please let us know.

rpavlik:
Closing due to no reply. The loader refactor in 1.0.6 should have fixed this, and if not that, some fixes I'm pulling in that are Android-related in an upcoming release.

PR #146, opened 2019-11-26
Single instance loader with forward compatibility

brycehutchings:
The loader currently has a design which makes it incompatible with extensions/core api newer than the loader for two reasons:

1. The loader keeps a handle map in order to map an input handle (first parameter to a function) into a dispatch table for a known instance. Every instance can potentially have a different dispatch table depending on which extensions and layers are enabled. If a future extension  function (accessed through xrGetInstanceProcAddr) creates a handle, the loader won't know about this and won't be able to look up a handle in the dispatch table map.
2. The loader's implementation of xrGetInstanceProcAddr uses a strict set of function names mapped to the dispatch table. If a function name is not a known one, it will fail.

My main goal is to make the loader forward compatible so that it can be used with extensions that come out after the loader has been built. One use case is so that complex engines with many separate components can work without coordinating which loader to package. A secondary goal is to shrink and simplify the loader to minimize the chance for it to contain bugs (which would ship with the app).

The high level changes are:

* LoaderInstance is no longer tracked through a per-handle map. Instead there is a ActiveLoaderInstance which allows only one outstanding XrInstance handle at a time. If xrCreateInstance is called when another XrInstance is still alive, it will fail with XR_ERROR_LIMIT_REACHED. For this reason, care must be taken not to leak XrInstance handles.
* xrGetInstanceProcAddr no longer maps function name to dispatch table entry. Instead it uses the "topmost" xrGetInstanceProcAddr. Topmost means the API layer closest to the application or the runtime if there are no API layers. In practice this looks like: app -> openxr_loader!xrGetInstanceProcAddr (trampoline) -> [0..N API layers!xrGetInstanceProcAddr] -> openxr_loader!LoaderXrTermGetInstanceProcAddr (terminator) -> runtime!xrGetInstanceProcAddr.
* Similarly, the way the dispatch table is generated now uses the same "topmost" xrGetInstanceProcAddr. Previously the loader will populate the dispatch table from the terminator+runtime, then iterate through the API layers in reverse allowing them to sparsely replace any function pointers that they supported (see LoaderInstance::CreateDispatchTable). This required a lot of generated code which is now gone. The only benefit that I could see to it was that if an API layer did not forward calls for unsupported functions then it would break the loader, but this is not a valid implementation. I did find the 'loader_test/test_layers' implementation, used for testing the loader, had this kind of bug (which I've fixed). All other layers implemented it correctly. Now the loader can re-use the GeneratedXrPopulateDispatchTable function to generate the dispatch table instead.
* The amount of Python code to generate loader code has shrunk from ~900 lines to ~350. It no longer automatically generates many of the dispatch table generating functions now that GeneratedXrPopulateDispatchTable is being reused with the "topmost" xrGetInstanceProcAddr. * The specification requires that xrGetInstanceProcAddr return XR_ERROR_FUNCTION_UNSUPPORTED for any function in an extension that is not enabled. Previously the loader did this check for all functions in the generated code. Now, the loader relies on the API layers and runtimes to enforce this. The loader will still do this for the debug utils extension because it supports it. In the spirit of being forward compatible, it is not possible for the loader to know which functions are in extensions that it does not know about, so I think this makes sense.

Design decisions:
* I have kept the implementation of LoaderInstance decoupled from the 'singleton' instance requirement in the loader.
* I kept kept the implementation of RuntimeInterface as-is, with its support for multiple instances. In practice though, it will only ever have a single instance in its maps.

In the future, if there is need by apps to have multiple simultaneous XrInstances, then an alternative loader design will be required in order to do this. So far from survey of those using the loader, we have not yet heard of a need for this though.

### DIFF
--- a/specification/loader/design.adoc
+++ b/specification/loader/design.adoc
@@ -242,12 +242,7 @@ static XrResult ApiLayerInterface::LoadApiLayers(
 ==== The LoaderInstance Class ====
 
 The primary OpenXR object is the `XrInstance`, and from that most other data
-is either queried or created.  In many ways, the `XrInstance` object is a
-loader object that manages all the other underlying OpenXR objects/handles.
-Because of this, a large amount of data needs to be associated with it by
-the loader.
-In order to do this, the loader maps an internal `LoaderInstance` object
-pointer to the returned `XrInstance` value.
+is either queried or created.
 
 A `LoaderInstance` is created during the OpenXR `xrCreateInstance` call, and
 destroyed during the `xrDestroyInstance` call.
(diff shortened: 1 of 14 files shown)


specification/loader/design.adoc
```diff
@@ -675,174 +608,14 @@ implemented in the loader.
...
-====
-* What do we do here?
-** I suggest we disallow mixing and matching using loader exports with
-   application generated dispatch tables.
-** At least where commands could take a different path for `xrCreate`
-   and usage.
-====
-
+The loader supports a single XrInstance at a time in order to avoid
+tracingk handle values and their relationship to the `LoaderInstance`. Every
```
rpavlik, line 615: typo: "tracingk"

rpavlik (approved):

rpavlik:
Merging prior to release - corresponding changes have been merged in internal source tree already.

PR #149, opened 2019-12-09
Link patches

rpavlik:
Some patches mostly about linking - things that came up in the Debian packaging.

Only vaguely bikeshed patch is the one renaming the runtime_list tool again - this just renames the output file to be `openxr_runtime_list` so that it can actually be installed (ala `glxinfo`, `vkinfo`)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -21,6 +21,7 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 ### Dependencies
+set(OpenGL_GL_PREFERENCE GLVND)
 find_package(OpenGL)
 if(OPENGL_FOUND)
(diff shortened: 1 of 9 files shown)


rpavlik:
For easier review, here's the diff without the (now) last 2 patches which are more formatting than actual change: https://github.com/KhronosGroup/OpenXR-SDK-Source/pull/149/files/e32a4c2ae393720468465816ae20366088c65d35..8ada0f8d3200cab6433b8ad6626ddedc0b95c581

rpavlik:
There, satisfied CI :)

rpavlik:
@ChristophHaag can you sanity check this patch please?

rpavlik:
Found and fixed one small issue with this when working on Debian packages.

ChristophHaag:
That's quite a lot of changed. I didn't look too closely but nothing stood out as being obviously wrong.

On Archlinux the linked libraries look good and everything still seems to work.

rpavlik:
Thanks for the check.

PR #155, opened 2019-12-20
Fix ApiLayerInterface::LoadApiLayers() failing to load available and enabled API layer

ghost:
In src/loader/api_layer_interface.cpp, when ApiLayerInterface::LoadApiLayers() is called, from api_layer_interface.cpp:197 up until api_layer_interface.cpp:234,
the OpenXR Loader constructs a list of API layers in the list enabled_api_layers.
I have not analyzed this section of code closely, but **in my case, this section of code adds XR_APILAYER_LUNARG_core_validation to the enabled_api_layers list twice.**

This results in the following bug when the below code is executed:
```
for (uint32_t layer = 0; layer < enabled_api_layers.size(); ++layer) {
    if (enabled_api_layers[layer] == manifest_file->LayerName()) {
    layer_found[layer] = true;
    enabled = true; 
    break;
  }
}
```
The **break** statement at the end of the for loop guarantees that a second API layer in the enabled_api_layers list of the exact same name as an API layer earlier in the list will never have its layer_found[layer] value set to "true", thus triggering the following error:
```
Error [GENERAL | xrCreateInstance | OpenXR-Loader] : ApiLayerInterface::LoadApiLayers - failed to find layer XR_APILAYER_LUNARG_core_validation
```

A hacky way of fixing this is by simply commenting out the break statement, and this seems to work fine with no side effects. (**This pull request simply removes the break statement.**)

A more proper fix might involve checking any API layer name against the current list of enabled_api_layers before adding it to the enabled_api_layers list.

### DIFF
--- a/src/loader/api_layer_interface.cpp
+++ b/src/loader/api_layer_interface.cpp
@@ -245,7 +245,6 @@ XrResult ApiLayerInterface::LoadApiLayers(const std::string& openxr_command, uin
                 if (enabled_api_layers[layer] == manifest_file->LayerName()) {
                     layer_found[layer] = true;
                     enabled = true;
-                    break;
                 }
             }
         }


rpavlik:
Did you make any progress on figuring out the "correct" fix here?

ghost:
@rpavlik No, I have not worked on or investigated this issue further.

rpavlik:
OK. The issue was because it was being enabled in two places, right? I'm happy with merging this as long as we have an issue filed for the underlying problem.

ghost:
Not sure if there's an issue ticket for this opened anywhere separate from this PR,
but yes, the issue was a combination of:
1. The API layer was enabled in two places, and
2. The code for checking whether a layer is found (the for loop mentioned in this PR) will never check if the second copy of the enabled API layer is found due to the break statement.

Thus the second mention of the API layer's layer_found value defaults to false, generating an error about the API layer not being found.

rpavlik:
Now there is one :) Thanks for the patch!

PR #157, opened 2020-01-02
hello_xr: Initialize hand_scale to 1.0

ChristophHaag:
Hand scale was default initialized to 0.0 which made controller cubes
invisible if the grabAction was not available/active on startup.


rpavlik:
Not sure why the CI is failing, though it looks like it wants clang-format applied.

PR #168, opened 2020-03-09
Fix loader test

### LINKED ISSUE #167: loader test failure
I've been trying to figure out why I persistently get this issue in the loader test: 

```
    Starting TestEnumLayers
        No explicit layers layer count check: Passed: 2 layers available.
        No explicit layers layer props query: Passed
           - XR_APILAYER_LUNARG_api_dump
           - XR_APILAYER_LUNARG_core_validation
        Simple explicit layers layer count check: Failed, expected count 8 (2 seen earlier plus 6 we added), got 6
```

After enabling more debug output, etc. I discovered that it's actually only enumerating the json files in the XR_API_LAYER_PATH environment var: that is, when XR_API_LAYER_PATH is set, it's not looking in the system environment variable anymore. (This appears to be correct according to the loader spec, anyway.)

Is this actually a bug in the test?

rpavlik:
Fixes #167 and makes the dummy runtime a little bit more useful - so you can run the openxr_runtime_list app against it.

### DIFF
--- a/src/scripts/generate_runtime_manifest.py
+++ b/src/scripts/generate_runtime_manifest.py
@@ -45,6 +45,19 @@ def main(argv):
         elif opt in ("-b", "--bad"):
             generate_badjson_jsons = True
 
+    file_text  = '{\n'
+    file_text += '    "file_format_version": "%s",\n' % cur_runtime_json_version
+    file_text += '    "runtime": {\n'
+    file_text += '        "library_path": "%s",\n' % library_location
+    file_text += '        "functions": {\n'
+    file_text += '           "xrNegotiateLoaderRuntimeInterface":\n'
+    file_text += '               "xrNegotiateLoaderRuntimeInterface"\n'
+    file_text += '       }\n'
+    file_text += '    }\n'
+    file_text += '}\n'
+    with open(output_file, 'w') as f:
+        f.write(file_text)
+
     if generate_badjson_jsons:
         # Bad File format versions
         ####################################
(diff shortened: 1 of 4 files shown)


src/scripts/generate_runtime_manifest.py
```diff
@@ -45,6 +45,19 @@ def main(argv):
         elif opt in ("-b", "--bad"):
             generate_badjson_jsons = True
 
+    file_text  = '{\n'
```
bradgrantham-lunarg, line 48: This seems like a great idea.  Should this be only if `not generate_bad_jsons`?

bradgrantham-lunarg (approved):
Approved but please consider my one comment about the python generator.  Thanks!

src/scripts/generate_runtime_manifest.py
rpavlik: So I don't think so, because first you need that for the dependency in CMake. But also, my goal here is actually to run the "list" app in a test suite, on its own and with layers, on top of this dummy runtime. So I need one manifest that is actually for a reasonably-valid entry point :)

bradgrantham-lunarg: It's just that you're not generating a bad json, which is what `-b` requests.

rpavlik: Well, we are actually generating the bad json. We're just also generating a good json.

If you'd rather not, I can revert this and just keep it in the Debian package.

bradgrantham-lunarg: I think I just needed to note my opinion.  Thanks, carry on!

PR #169, opened 2020-03-09
Manpages

rpavlik:
This adds a manpage for the two binaries, written using the (admittedly BSD-centric, but works on Linux as well) mandoc "semantic" troff markup. Right now it should install in the right place on anything except Windows.

Requesting review from @ChristophHaag since he's the only other packager of OpenXR I know :)


Wallbraker:
Looks good to me! :+1:

PR #171, opened 2020-03-21
Consistent bitflag naming

Ralith:
These broke my binding generator.


rpavlik:
Good catch, thanks. I've filed an issue up on OpenXR-Docs to track this (since there would be spec changes to match the registry changes, and because my issue-sync script doesn't sync pull requests to the internal tracker, only issues), I'm hopeful we can get this resolved since the whole thing is provisional.

rpavlik:
These have been fixed on all repositories with the 1.0.8 release. Thanks for the report! 

(PS. We now have a script in CI making sure this doesn't happen again.)

PR #173, opened 2020-03-30
Remove static initializers from loader

chrdavis:
This change replaces global static initializers with functions that return static locals. With this change code that includes OpenXR doesn't have to page in this code and initialize these during startup.

### DIFF
--- a/src/loader/loader_core.cpp
+++ b/src/loader/loader_core.cpp
@@ -664,7 +664,7 @@ XRAPI_ATTR XrResult XRAPI_CALL LoaderXrTermCreateDebugUtilsMessengerEXT(XrInstan
         *messenger = reinterpret_cast<XrDebugUtilsMessengerEXT>(temp_mess_ptr);
     }
     if (XR_SUCCEEDED(result)) {
-        LoaderLogger::GetInstance().AddLogRecorder(MakeDebugUtilsLoaderLogRecorder(createInfo, *messenger));
+        LoaderLogger::GetInstance().AddLogRecorderForXrInstance(instance, MakeDebugUtilsLoaderLogRecorder(createInfo, *messenger));
         RuntimeInterface::GetRuntime().TrackDebugMessenger(instance, *messenger);
     }
     LoaderLogger::LogVerboseMessage("xrCreateDebugUtilsMessengerEXT", "Completed loader terminator");
--- a/src/loader/loader_core.cpp
+++ b/src/loader/loader_core.cpp
@@ -44,11 +44,18 @@
 #include <string>
 #include <utility>
 #include <vector>
+
 // Global lock to prevent reading JSON manifest files at the same time.
-static std::mutex g_loader_json_mutex;
+std::mutex &GetLoaderJsonMutex() {
+    static std::mutex loader_json_mutex;
+    return loader_json_mutex;
+}
 
 // Global lock to prevent simultaneous instance creation/destruction
-static std::mutex g_instance_create_destroy_mutex;
+std::mutex &GetInstanceCreateDestroyMutex() {
+    static std::mutex instance_create_destroy_mutex;
+    return instance_create_destroy_mutex;
+}
 
 // Terminal functions needed by xrCreateInstance.
 XRAPI_ATTR XrResult XRAPI_CALL LoaderXrTermGetInstanceProcAddr(XrInstance, const char *, PFN_xrVoidFunction *);
@@ -84,7 +91,7 @@ LOADER_EXPORT XRAPI_ATTR XrResult XRAPI_CALL xrEnumerateApiLayerProperties(uint3
     LoaderLogger::LogVerboseMessage("xrEnumerateApiLayerProperties", "Entering loader trampoline");
 
     // Make sure only one thread is attempting to read the JSON files at a time.
-    std::unique_lock<std::mutex> json_lock(g_loader_json_mutex);
+    std::unique_lock<std::mutex> json_lock(GetLoaderJsonMutex());
 
     XrResult result = ApiLayerInterface::GetApiLayerProperties("xrEnumerateApiLayerProperties", propertyCapacityInput,
                                                                propertyCountOutput, properties);
@@ -118,7 +125,7 @@ xrEnumerateInstanceExtensionProperties(const char *layerName, uint32_t propertyC
 
     {
         // Make sure only one thread is attempting to read the JSON files at a time.
-        std::unique_lock<std::mutex> json_lock(g_loader_json_mutex);
+        std::unique_lock<std::mutex> json_lock(GetLoaderJsonMutex());
 
         // Get the layer extension properties
         result = ApiLayerInterface::GetInstanceExtensionProperties("xrEnumerateInstanceExtensionProperties", layerName,
@@ -235,7 +242,7 @@ LOADER_EXPORT XRAPI_ATTR XrResult XRAPI_CALL xrCreateInstance(const XrInstanceCr
         return XR_ERROR_VALIDATION_FAILURE;
     }
 
-    std::unique_lock<std::mutex> instance_lock(g_instance_create_destroy_mutex);
+    std::unique_lock<std::mutex> instance_lock(GetInstanceCreateDestroyMutex());
 
     // Check if there is already an XrInstance that is alive. If so, another instance cannot be created.
     // The loader does not support multiple simultaneous instances because the loader is intended to be
@@ -252,7 +259,7 @@ LOADER_EXPORT XRAPI_ATTR XrResult XRAPI_CALL xrCreateInstance(const XrInstanceCr
 
     // Make sure only one thread is attempting to read the JSON files and use the instance.
     {
-        std::unique_lock<std::mutex> json_lock(g_loader_json_mutex);
+        std::unique_lock<std::mutex> json_lock(GetLoaderJsonMutex());
         // Load the available runtime
         result = RuntimeInterface::LoadRuntime("xrCreateInstance");
         if (XR_FAILED(result)) {
@@ -329,7 +336,7 @@ LOADER_EXPORT XRAPI_ATTR XrResult XRAPI_CALL xrDestroyInstance(XrInstance instan
         return XR_ERROR_HANDLE_INVALID;
     }
 
-    std::unique_lock<std::mutex> loader_instance_lock(g_instance_create_destroy_mutex);
+    std::unique_lock<std::mutex> loader_instance_lock(GetInstanceCreateDestroyMutex());
 
     LoaderInstance *loader_instance;
     XrResult result = ActiveLoaderInstance::Get(&loader_instance, "xrDestroyInstance");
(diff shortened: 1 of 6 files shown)


bradgrantham-lunarg:
Your edits look fine, but this seems equivalent to what was there.  Help me understand why this is necessary - are the bare C++ initializers not honored during DSO load or some such?

chrdavis:
> Your edits look fine, but this seems equivalent to what was there. Help me understand why this is necessary - are the bare C++ initializers not honored during DSO load or some such?

global static initializers are initialized at module load time.  This results in code being paged in that most likely isn't necessary contributing to performance regressions.  The change is to move these to local statics that are initialized on demand.

See:
https://chromium.googlesource.com/chromium/src.git/+/master/docs/static_initializers.md

src/loader/loader_core.cpp
```diff
@@ -657,7 +664,7 @@ XRAPI_ATTR XrResult XRAPI_CALL LoaderXrTermCreateDebugUtilsMessengerEXT(XrInstan
         *messenger = reinterpret_cast<XrDebugUtilsMessengerEXT>(temp_mess_ptr);
     }
     if (XR_SUCCEEDED(result)) {
-        LoaderLogger::GetInstance().AddLogRecorderForXrInstance(instance, MakeDebugUtilsLoaderLogRecorder(createInfo, *messenger));
+        LoaderLogger::GetInstance().AddLogRecorder(MakeDebugUtilsLoaderLogRecorder(createInfo, *messenger));
```
brycehutchings: Why is the logger no longer associated with the instance handle?

bradgrantham-lunarg: Yeah, what Bryce said. :)

chrdavis: Thanks.  Looks like a typo I missed when moving to this repo.

bradgrantham-lunarg:
> global static initializers are initialized at module load time. This results in code being paged in that most likely isn't necessary contributing to performance regressions. The change is to move these to local statics that are initialized on demand.
> 
> See:
> https://chromium.googlesource.com/chromium/src.git/+/master/docs/static_initializers.md

Okay, that helps.  Makes sense, thanks.

I would think (based on an initial scan) all these initializers will be called before any work is done in an api or runtime, so I can't imagine this changing the performance characteristics significantly if this is dynamically loaded.  Is this change required before Chromium will allow loading the loader DSO?

bradgrantham-lunarg:
> I would think (based on an initial scan) all these initializers will be called before any work is done in an api or runtime, so I can't imagine this changing the performance characteristics significantly if this is dynamically loaded. Is this change required before Chromium will allow loading the loader DSO?

(To be clear, I don't really have a problem with using this style, the bare initializers are a little awkward.  However, I'm against making changes just to match another project's style just because all changes always bring a risk of regression.  I don't object though and I'll kick it back to Ryan.)

chrdavis:
> > global static initializers are initialized at module load time. This results in code being paged in that most likely isn't necessary contributing to performance regressions. The change is to move these to local statics that are initialized on demand.
> > See:
> > https://chromium.googlesource.com/chromium/src.git/+/master/docs/static_initializers.md
> 
> Okay, that helps. Makes sense, thanks.
> 
> I would think (based on an initial scan) all these initializers will be called before any work is done in OpenXR, so I can't imagine this changing the performance characteristics significantly. Is this change required before Chromium will allow loading the loader DSO?

This would impact chromium startup - particularly on spinning disks.  This was a few kb of the multi-MB cleanup of static initializers that occur on chrome.dll load.  The point is that chromium pays these costs whether we use the OpenXR codebase or not.

brycehutchings:
> > > global static initializers are initialized at module load time. This results in code being paged in that most likely isn't necessary contributing to performance regressions. The change is to move these to local statics that are initialized on demand.
> > > See:
> > > https://chromium.googlesource.com/chromium/src.git/+/master/docs/static_initializers.md
> > 
> > 
> > Okay, that helps. Makes sense, thanks.
> > I would think (based on an initial scan) all these initializers will be called before any work is done in OpenXR, so I can't imagine this changing the performance characteristics significantly. Is this change required before Chromium will allow loading the loader DSO?
> 
> This would impact chromium startup - particularly on spinning disks. This was a few kb of the multi-MB cleanup of static initializers that occur on chrome.dll load. The point is that chromium pays these costs whether we use the OpenXR codebase or not.

Is delay load or dynamic load an option? I would imagine loading the loader dynamically would bring even bigger improvements since the DLL (and its dependencies) won't load at all except when needed.

[EDIT: Just realized the loader is probably static linked]

chrdavis:
> > > > global static initializers are initialized at module load time. This results in code being paged in that most likely isn't necessary contributing to performance regressions. The change is to move these to local statics that are initialized on demand.
> > > > See:
> > > > https://chromium.googlesource.com/chromium/src.git/+/master/docs/static_initializers.md
> > > 
> > > 
> > > Okay, that helps. Makes sense, thanks.
> > > I would think (based on an initial scan) all these initializers will be called before any work is done in OpenXR, so I can't imagine this changing the performance characteristics significantly. Is this change required before Chromium will allow loading the loader DSO?
> > 
> > 
> > This would impact chromium startup - particularly on spinning disks. This was a few kb of the multi-MB cleanup of static initializers that occur on chrome.dll load. The point is that chromium pays these costs whether we use the OpenXR codebase or not.
> 
> Is delay load or dynamic load an option? I would imagine loading the loader dynamically would bring even bigger improvements since the DLL (and its dependencies) won't load at all except when needed.
> 
> [EDIT: Just realized the loader is probably static linked]

Correct.  Also, if the chrome process is sandboxed we can't realize a delay loaded module :(.

bradgrantham-lunarg:
> Correct. Also, if the chrome process is sandboxed we can't realize a delay loaded module :(.

... How would you load API layer DLLs and runtime DLLs?

[edit: ^DLL^shared objects specific to platform^ ]

chrdavis:
> > Correct. Also, if the chrome process is sandboxed we can't realize a delay loaded module :(.
> 
> ... How would you load API layer DLLs and runtime DLLs?
> 
> [edit: ^DLL^shared objects specific to platform^ ]

Unfortunately for this reason there is a list of modules we cannot delay load for chromium binaries on Windows.  I've done a considerable amount of work for the next release of Win10 to decrease the impact of the dependencies of those dependencies, though.

bradgrantham-lunarg:
I suppose last comment is both for my edification and also for a product.

1) The loader is intended to be dynamically loaded as well as load other objects.  Again, your change seems fine, but I'm wondering whether backstory is important.

2) Pluto VR and LunarG are in the process of open sourcing (hopefully today) an API layer providing an extension.  Will Chromium's operation require us to work with you to include API layers like that statically, or will there be a way for users to set API layers for OpenXR's loader to load?  That seems painful (but I can understand it might be necessary).

chrdavis:
> I suppose last comment is both for my edification and also for a product.
> 
> 1. The loader is intended to be dynamically loaded as well as load other objects.  Again, your change seems fine, but I'm wondering whether backstory is important.
> 2. Pluto VR and LunarG are in the process of open sourcing (hopefully today) an API layer providing an extension.  Will Chromium's operation require us to work with you to include API layers like that statically, or will there be a way for users to set API layers for OpenXR's loader to load?  That seems painful (but I can understand it might be necessary).

Sorry I can't speak for that.  I'm just fixing some perf bugs all across the codebase.

bradgrantham-lunarg:
> > > Correct. Also, if the chrome process is sandboxed we can't realize a delay loaded module :(.
> > 
> > 
> > ... How would you load API layer DLLs and runtime DLLs?
> > [edit: ^DLL^shared objects specific to platform^ ]
> 
> Unfortunately for this reason there is a list of modules we cannot delay load for chromium binaries on Windows. I've done a considerable amount of work for the next release of Win10 to decrease the impact of the dependencies of those dependencies, though.

Sorry, could you clarify, if I want to load a brand-new OpenXR runtime in my browser (for, what, WebXR, I guess?) how would I do that?

@rpavlik @brycehutchings Should this be an issue, or was it already raised as an issue?  This makes "desktop" OpenXR operation difficult for this execution environment.  Like could Monado's DSO be loaded as a runtime or would Google/Chromium become the gatekeeper for "approved" API layers and runtimes?  This seems less than ideal.

chrdavis:
> > > > Correct. Also, if the chrome process is sandboxed we can't realize a delay loaded module :(.
> > > 
> > > 
> > > ... How would you load API layer DLLs and runtime DLLs?
> > > [edit: ^DLL^shared objects specific to platform^ ]
> > 
> > 
> > Unfortunately for this reason there is a list of modules we cannot delay load for chromium binaries on Windows. I've done a considerable amount of work for the next release of Win10 to decrease the impact of the dependencies of those dependencies, though.
> 
> Sorry, could you clarify, if I want to load a brand-new OpenXR runtime in my browser (for, what, WebXR, I guess?) how would I do that?
> 
> @rpavlik @brycehutchings Should this be an issue, or was it already raised as an issue? This makes "desktop" OpenXR operation difficult for this execution environment. Like could Monado's DSO be loaded as a runtime or would Google/Chromium become the gatekeeper for "approved" API layers and runtimes? This seems less than ideal.

I feel like this has been taken way out of context.  Nothing has changed in how chromium sandboxing works.  I don't have the background on what you are worried about - but if it currently isn't broken it isn't an issue.

bradgrantham-lunarg (approved):
The changes seem fine to me.  But it sounds like if there's a policy around loading dynamic modules in Chromium this opens a can of worms with relation to users choosing which API layers and runtimes they want, so should precipitate an additional issue with regards to cooperation and communication with Chromium

chrdavis:
> The changes seem fine to me. But it sounds like if there's a policy around loading dynamic modules in Chromium this opens a can of worms with relation to users choosing which API layers and runtimes they want, so should precipitate an additional issue with regards to cooperation and communication with Chromium

Again - this is way out of context.  I'm just a Chromium contributor.  Sandboxing in chromium has been around for years.  There is no "policy".

bradgrantham-lunarg:
> > Sorry, could you clarify, if I want to load a brand-new OpenXR runtime in my browser (for, what, WebXR, I guess?) how would I do that?
> > @rpavlik @brycehutchings Should this be an issue, or was it already raised as an issue? This makes "desktop" OpenXR operation difficult for this execution environment. Like could Monado's DSO be loaded as a runtime or would Google/Chromium become the gatekeeper for "approved" API layers and runtimes? This seems less than ideal.
> 
> I feel like this has been taken way out of context. Nothing has changed in how chromium sandboxing works. I don't have the background on what you are worried about - but if it currently isn't broken it isn't an issue.

Yeah, I agree your changes don't break things.  But it introduces a concern in which I'm referencing Ryan and Bryce because OpenXR may want to have a conversation as a standards body around the loader, layers, and runtimes.

[quick edit: I did approve, and will try to take conversation about sandboxing offline]

brycehutchings (approved):

chrdavis:
Did the build verification automation stall?

rpavlik:
GitHub just lost track of it - if you click through, it was all successful.

chrdavis:
Can somebody merge this PR?

rpavlik:
Ah, yes, sorry I missed the reviews.

PR #175, opened 2020-04-08
Build pipeline: Remove static libraries, and build arm+arm64 along with UWP

brycehutchings:
We (Microsoft) needed the loader built for UWP packages and ARM/ARM64 so this modifies the build a bit to support that. Summary of changes and why:
1. "UWP" target was added because HoloLens 2 apps will need them.
2. Removed builds of the static loader. I do not believe it is safe to build static libs because compiler settings such as around exception handling can impact whether things link together (to the best of my knowledge at least).
3. Since the DLL is a robust ABI, we don't need to draw attention to msvs2019 as opposed to msvs2017 so I removed that from the directory names
4. I re-arranged the final output artifact because it assumed only 32-bit and 64-bit Intel, but we need ARM and ARM64 too. I changed the root folder names so that they include the target platform name and then each folder contains a 'lib' and 'bin'  subfolder (no more lib32/bin32). I also moved the 'include' folder to the root because every copy is the same.

The removal of static lib might be debatable so if this is not desired I can undo it. I'm also happy to change the final folder structure. You can see my pipeline results here including the artifacts: https://bryceh.visualstudio.com/OpenXR/_build/results?buildId=26&view=results


rpavlik:
Can I trouble you for a changelog fragment?

rpavlik (approved):

PR #178, opened 2020-04-24
Keep runtime loaded after xrEnumerateInstanceExtensionProperties call

brycehutchings:
Most apps follow the pattern of calling xrEnumerateInstanceExtensionProperties before xrCreateInstance to inspect available extensions. This type of app will end up loading the runtime three times (and unloading two times) on startup:
1. First call to xrEnumerateInstanceExtensionProperties (two-call idiom, get the count) causes runtime to load an unload.
2. Second call to xrEnumerateInstanceExtensionProperties (two-call idiom, fill the buffer) causes the runtime to load and unload
3. xrCreateInstance causes the runtime to load

This is wildly inefficient. This change will result in the runtime staying alive after the first call to xrEnumerateInstanceExtensionProperties so that the runtime only loads once. The runtime only unloads if xrCreateInstance fails at a late stage (after the runtime was loaded) or on xrDestroyInstance.

The only scenario I can think this would affect is an app that only calls xrEnumerateInstanceExtensionProperties and never calls xrCreateInstance. In the case this type of app stays alive for a long time and the active runtime changes, the loader will not honor the change if an instance is then created. I do not know of an app that behaves this way (except for loader_tests). If this is ever needed, the loader could expose a loader-only function such as xrUnloadRuntime as an escape hatch.

### DIFF
--- a/specification/loader/design.adoc
+++ b/specification/loader/design.adoc
@@ -50,11 +50,8 @@ call if it requires a list of available runtime manifest files.
 [source,c++]
 ----
 static XrResult RuntimeManifestFile::FindManifestFiles(
-            ManifestFileType type,
             std::vector<std::unique_ptr<RuntimeManifestFile>> &manifest_files);
 ----
-  * pname:type indicates the specific type of manifest file being searched for.
-    In this case, it must: be `MANIFEST_TYPE_RUNTIME`.
   * pname:manifest_files is a vector that will be used to store a unique_ptr to
     each valid runtime manifest file found.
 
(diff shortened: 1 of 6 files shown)


rpavlik:
Can you add a changelog fragment? It looks reasonable to me, though I'm not among the most stringent in my requirements for shared libraries.

(I assume you tried this with the WIP conformance tests and verified it isn't leaking?)

bradgrantham-lunarg:
Dave Houlton no longer works for LunarG and as far as I know is not involved in OpenXR or Khronos work.

rpavlik:
Unfortunately GitHub doesn't read this closely so it keeps suggesting him

rpavlik:
@brycehutchings I suspect this is already merged from gitlab - am I right?

brycehutchings:
Yup I will close.

PR #181, opened 2020-04-29
hello_xr/vulkan: Fix build with Vulkan headers 1.2.136.

### LINKED ISSUE #180: Build broken with latest Vulkan headers
cc @oddhack 

Looks like `VK_DEBUG_REPORT_OBJECT_TYPE_OBJECT_TABLE_NVX_EXT` got removed, and while OpenXR didn't use it directly, it was mentioned in an enum list:

```
_(OBJECT_TABLE_NVX)
```

Not sure if this counts as Vulkan breakage, but I suspect the right solution for now is for us to just remove all X-suffixed entries from those enum lists. (Ideally Vulkan would have something like our openxr_reflection.h header so no app-made list like this would be required.)

See also:

- Bug report in Debian https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=958837
- Patch in ArchLinux: https://aur.archlinux.org/cgit/aur.git/tree/0001-hello_xr-vulkan-Fix-build-with-Vulkan-headers-1.2.13.patch?h=openxr-loader-git

rpavlik:
Patch from @lubosz

Fixes #180

### DIFF
--- a/changes/sdk/pr.181.OpenXR-SDK-Source.md
+++ b/changes/sdk/pr.181.OpenXR-SDK-Source.md
@@ -1,0 +1,4 @@
+---
+- issue.180.OpenXR-SDK-Source
+---
+hello_xr: Fix build with Vulkan headers 1.2.136.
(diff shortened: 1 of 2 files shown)


rpavlik:
@lubosz if you could do the CLA when you get a chance, that would be great, thanks.

rpavlik:
thank you @lubosz !

PR #188, opened 2020-05-06
cmake: Check for C++-17 and conditionally enable it

philn:
In GCC __cplusplus will be 201703 only if the -std=c++17 option is passed to the
compiler. Without this option the filesystem_utils.cpp won't compile under GCC
9.3.0, giving the following error:

src/common/filesystem_utils.cpp:74:10: fatal error: experimental/filesystem: No
such file or directory

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -22,6 +22,15 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
+include(CheckCXXCompilerFlag)
+
+if("${CMAKE_MAJOR_VERSION}" EQUAL 3 AND "${CMAKE_MINOR_VERSION}" GREATER 7)
+    check_cxx_compiler_flag(-std=c++17 HAVE_FLAG_STD_CXX17)
+    if(HAVE_FLAG_STD_CXX17)
+        set(CMAKE_CXX_STANDARD 17)
+    endif()
+endif()
+
 include(GNUInstallDirs)
 
 ### Dependencies
--- a/src/loader/CMakeLists.txt
+++ b/src/loader/CMakeLists.txt
@@ -143,11 +143,19 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     endif()
 
     set_target_properties(openxr_loader PROPERTIES SOVERSION "${MAJOR}" VERSION "${MAJOR}.${MINOR}.${PATCH}")
-    target_link_libraries(
-        openxr_loader
-        PRIVATE stdc++fs
-        PUBLIC m
-    )
+
+    if(HAVE_FLAG_STD_CXX17)
+        target_link_libraries(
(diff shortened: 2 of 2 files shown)


rpavlik:
Ah, so they have removed the experimental headers in 9.3.0? In the older versions of GCC (and libstdc++) I've used, the C++14 setting that's on now is enough to get the experimental filesystem support. I'll look closer at this soon.

philn:
From what I can see here, the `filesystem` header is indeed no longer in `experimental/`.

rpavlik:
Thanks for doing this! Can you test and see if the branch in #192 works for you? I ended up making a fairly dramatic change, and actually checking to see if we could build using experimental/filesystem and filesystem (and filesystem with c++17 enabled), as well as then checking if we needed stdc++fs (since I can build c++17 mode here on Debian Buster with 8.3, but still need stdc++fs) to make it compile.

src/CMakeLists.txt
```diff
@@ -22,6 +22,15 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
+include(CheckCXXCompilerFlag)
+
+if("${CMAKE_MAJOR_VERSION}" EQUAL 3 AND "${CMAKE_MINOR_VERSION}" GREATER 7)
+    check_cxx_compiler_flag(-std=c++17 HAVE_FLAG_STD_CXX17)
```
rpavlik, line 28: I think the best way to do this that I've found is to search `CMAKE_CXX_COMPILE_FEATURES` for `cxx_std_17`:

```
list(FIND CMAKE_CXX_COMPILE_FEATURES cxx_std_17 cxx17_index)
if(NOT ${cxx17_index} EQUAL -1)
    set(CMAKE_CXX_STANDARD 17)
endif()
```

This currently will not behave as desired on MSVC.

This change could be rebased on top of #192 - it's no longer needed to make it build on gcc 9.3.0 (I hope!) but it would still be worth considering an "opportunistic" upgrade to C++17.

src/loader/CMakeLists.txt
```diff
@@ -143,11 +143,19 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
...
     set_target_properties(openxr_loader PROPERTIES SOVERSION "${MAJOR}" VERSION "${MAJOR}.${MINOR}.${PATCH}")
-    target_link_libraries(
-        openxr_loader
-        PRIVATE stdc++fs
-        PUBLIC m
-    )
+
+    if(HAVE_FLAG_STD_CXX17)
+        target_link_libraries(
+            openxr_loader
```
rpavlik, line 149: So this could be split into two `target_link_libraries` calls - one with just m, and one with stdc++fs. However, m isn't apparently actually needed, and at least on my gcc 8.3.0, I need stdc++fs even though I have c++17 support. I'd suggest just dropping this part of the patch when rebasing on #192 since that handles it in a more complete way.

rpavlik (changes requested):

rpavlik:
Closing because #192 was merged and was reported to fix the issue.

Issue #185, opened 2020-05-04
Export of CMake config-file package missing during install


matthieucoquet:
Installing the `openxr_loader` target should also export a `OpenXRConfig.cmake` file. An application using CMake could easily import it using `find_package()` (without writing a `FindOpenXR.cmake`).

If `openxr_loader` was build and installed using the default `CMAKE_INSTALL_PREFIX` (C:/Program Files (x86)/OPENXR on Windows and /usr/local on Linux). Then the application's CMake just need: 
```
find_package(OpenXR CONFIG REQUIRED)
target_link_libraries(app_name PRIVATE xr::openxr_loader)
```

The user doesn't need to add environment variables. CMake already searches in the correct folder.

However, if the user chooses to change the `CMAKE_INSTALL_PREFIX`. Then it just need to set `OpenXR_ROOT` environment or CMake variable before running the application's CMake. 

You can check CMake package documentation for more info.
https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html
https://cmake.org/cmake/help/latest/command/find_package.html

I made a first commit in my fork that is working on Windows, but I can't test on Linux right now. It's exporting only the target `openxr_loader`, and I'm not sure if other targets need to be exported.
https://github.com/matthieucoquet/OpenXR-SDK-Source/commit/6f50b3c0c10b103c4afe68b6ad38fd17578ebcb0

rpavlik:
Thanks for filing this! It was at the back of my mind but I hadn't gotten around to this yet - though I had made a Find module. Would need to make sure it works with our generated binary packages. Also, would be best to modify the library namespace from `xr` to `OpenXR` to match the module name.

matthieucoquet:
> Would need to make sure it works with our generated binary packages.

I'n not sure what binary packages you're referring to. Are you talking about the generated OpenXR-SDK?

I made some small changes to my fork:
* I renamed the namespace to OpenXR
* I added a OpenXR::headers target to only import the headers

So right now, two targets are imported: OpenXR::openxr_loader and OpenXR::headers

I plan to test this on Linux next week.

rpavlik:
ah, great! Feel free to submit a PR.

(Yes those are the packages I'm talking about. I can probably handle adding the extra CMake-foo to deal with them.)

rpavlik:
The basics of this were in 1.0.10, thanks for the contribution. I'll write the little helper needed to deal with the packages

rpavlik:
[...]

PR #191, opened 2020-05-13
Export of CMake config-file package during install

matthieucoquet:
See https://github.com/KhronosGroup/OpenXR-SDK-Source/issues/185

I tested on Windows 10 and Ubuntu 20.04 without changing the default CMake installation directory.

I could import it from another application with:
```
find_package(OpenXR CONFIG REQUIRED)
target_link_libraries(app_name PRIVATE OpenXR::openxr_loader)
```


rpavlik:
Looks pretty good to me! I've asked for someone else to pitch in a review as well (mostly just a test).  (Thanks for doing the changelog fragment, btw!)

rpavlik (approved):

PR #193, opened 2020-05-28
Missing attributes

rpavlik:
Building layers on some non-x86 platforms highlighted these were missing.  Fixes required for Android to even build the layers on arm-ish things. I'd feel better with a review from somebody with a HoloLens 2 before merging.

### DIFF
--- a/src/scripts/api_dump_generator.py
+++ b/src/scripts/api_dump_generator.py
@@ -119,15 +119,15 @@ class ApiDumpOutputGenerator(AutomaticSourceOutputGenerator):
     #   self            the ApiDumpOutputGenerator object
     def outputLayerHeaderPrototypes(self):
         generated_prototypes = '// Layer\'s xrGetInstanceProcAddr\n'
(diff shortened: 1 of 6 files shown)


src/scripts/api_dump_generator.py
```diff
@@ -1080,7 +1080,7 @@ def outputLayerCommands(self):
                     generated_commands += '#if %s\n' % cur_cmd.protect_string
 
                 prototype = cur_cmd.cdecl.replace(" xr", " ApiDumpLayerXr")
-                prototype = prototype.replace(self.genOpts.apicall, "").replace(self.genOpts.apientry, "")
+                # prototype = prototype.replace(self.genOpts.apicall, "").replace(self.genOpts.apientry, "")
```
brycehutchings: did you mean to comment this out?

brycehutchings: Seems like this line should just be removed?

rpavlik: Yeah should just remove it, good catch.

brycehutchings (approved):

PR #196, opened 2020-07-23
Script to build NuGet package from OpenXR-SDK Release

brycehutchings:
This script can be run to generate the official NuGet package. It takes the released .zip file that is generated in the OpenXR-SDK CI as input to generate the .nupkg file.


brycehutchings:
I have now automated it so the NuPkg is generated as part of the build and published as an artifact.

rpavlik (approved):
Looks great, thanks! Did you care to add a changelog fragment to explain what this adds?

PR #198, opened 2020-08-02
Work around std::filesystem::canonical bug on UWP

### LINKED ISSUE #197: Build fails under MSVC 2017
> OpenXR-SDK\src\common\filesystem_utils.cpp(219): error C3861: 'PathCanonicalizeW': identifier not found

Looks like this is supposed to come from Shlwapi.h ([docs](https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-pathcanonicalizew)) which isn't included anywhere that I can see.

brycehutchings:
We found a change in the 1.0.10 loader to use std::filesystem::canonical broke UWP apps on Windows (used mostly for HoloLens 2). Based on my investigation this STL function is broken on the UWP platform and always fails with an access denied error code. I will be opening a bug on the Microsoft STL about this. The use of canonical was introduced for Linux so that a relative path in a JSON manifest would be relative to the real manifest file and not the symbol link path. This functionality is not important on Windows because the registry is used for redirection instead, and so symbol links are not needed and their use is also uncommon on Windows. For this reason PathCchCanonicalize is used instead of std::filesystem::canonical on Windows. This also fixes #197 by replacing use of PathCanonicalize with PathCchCanonicalize which is documented as the recommended API.

### DIFF
--- a/src/common/filesystem_utils.cpp
+++ b/src/common/filesystem_utils.cpp
@@ -88,6 +88,10 @@
 #endif
 
 #if defined(XR_USE_PLATFORM_WIN32)
+#include <PathCch.h>
+#endif
+
+#if defined(XR_USE_PLATFORM_WIN32)
 #define PATH_SEPARATOR ';'
 #define DIRECTORY_SYMBOL '\\'
 #define ALTERNATE_DIRECTORY_SYMBOL '/'
@@ -128,7 +132,18 @@ bool FileSysUtilsGetAbsolutePath(const std::string& path, std::string& absolute)
 }
 
 bool FileSysUtilsGetCanonicalPath(const std::string& path, std::string& canonical) {
+#if defined(XR_USE_PLATFORM_WIN32)
+    // std::filesystem::canonical fails on UWP and must be avoided. This alternative will not
+    // follow symbolic links but symbolic links are not needed on Windows since the loader uses
+    // the registry as a form of indirection instead.
+    wchar_t canonical_wide_path[MAX_PATH];
+    if (FAILED(PathCchCanonicalize(canonical_wide_path, MAX_PATH, utf8_to_wide(path).c_str()))) {
+        return false;
+    }
+    canonical = wide_to_utf8(canonical_wide_path);
+#else
     canonical = FS_PREFIX::canonical(path).string();
+#endif
     return true;
 }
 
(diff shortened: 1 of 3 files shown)


rpavlik (changes requested):
Ah, good catch. I know the canonical check was tested on Windows with/without std::filesystem but I imagine the contributor didn't try building for HoloLens. Just add the changelog fragment and it'll be good to go!

brycehutchings:
Sorry I appear to be incapable of remembering to include the changelog fragment!

src/common/filesystem_utils.cpp
```diff
@@ -128,7 +132,18 @@ bool FileSysUtilsGetAbsolutePath(const std::string& path, std::string& absolute)
 }
 
 bool FileSysUtilsGetCanonicalPath(const std::string& path, std::string& canonical) {
+#if defined(XR_USE_PLATFORM_WIN32)
+    // std::filesystem::canonical fails on UWP and must be avoided. This alternative will not
+    // follow symbolic links but symbolic links are not needed on Windows since the loader uses
+    // the registry as a form of indirection instead.
+    wchar_t canonical_wide_path[MAX_PATH];
+    if (FAILED(PathCchCanonicalize(canonical_wide_path, MAX_PATH, utf8_to_wide(path).c_str()))) {
```
yl-msft, line 140: don't we add "W" at these functions to force them wchar version?

yl-msft (approved):

src/common/filesystem_utils.cpp
brycehutchings: yeah seems like a good idea!

brycehutchings: Actually PathCchCanonicalize is wide only

rpavlik, line 140: isn't max_path an artificial limitation when using wide apis now?

brycehutchings: It's not related to wide characters per-se. We can use a bigger number, but many parts of windows use MAX_PATH so it is a bit of an unfortunate de-facto limit in Windows. Another option is to use PathCchCanonicalizeEx and dynamically allocate the buffer to the length needed. What I did was a bit of a lazy but common pattern in Windows programming.

rpavlik: ah, ok. That will work until it doesn't. :)

rpavlik: there, filed an issue on it so we don't forget :)

chrisfromwork:
@rpavlik  do you know when this change will make it into the OpenXR-SDK repo?

rpavlik:
Next patch release at the latest. Should be very soon.

PR #199, opened 2020-08-04
Clean up WIN32/WindowsStore in CMake projects.

brycehutchings:
This simplifies things in most places by removing WindowsStore-specific
code or fixing areas that assume "Windows" system when they should
assume WIN32 (which includes WindowsStore). As a consequence, more
projects can be built for UWP. Finally, this fixes build issues when
building ARM/ARM64 for Win32 related to bad OpenGL and Vulkan
detection in CMake.


src/tests/list/CMakeLists.txt
```diff
@@ -41,8 +41,6 @@ if(MSVC)
 endif()
 
 set_target_properties(runtime_list PROPERTIES FOLDER ${TESTS_FOLDER})
-set_target_properties(runtime_list PROPERTIES OUTPUT_NAME openxr_runtime_list)
```
rpavlik, line 44: Why was this removed?

rpavlik (commented):
Looks pretty good, just have one question where a change I made got reverted. I had namespaced the name of the runtime list app so it could be installed system wide on Linux. If that breaks some other platform I can carry the patch in the Debian packaging, but I'd also like to know what gets broken 😁

src/tests/list/CMakeLists.txt
brycehutchings: Yeah it broke the UWP packaging support in CMake because it puts the entrypoint executable name in as runtime_list and then it can't find the file (I guess it isn't polished enough to use OUTPUT_NAME if specified). Can we rename the entire target to openxr_runtime_list instead?

rpavlik: Ah, weird, go figure. Yep, we can rename the whole thing, that works for me!

brycehutchings: Okay renamed the whole target

rpavlik:
awesome, thanks!

rpavlik (approved):

PR #206, opened 2020-09-06
Fix cmake target export when installing both Release and Debug config on Windows

matthieucoquet:
Before the fix, the library was always installed in `lib/openxr_loader.lib`. If we tried to installed both Release and Debug, one will overwrite the other, resulting in linking error during application compilation.

The fix is to install the debug lib to another file. With this fix, the debug static lib will be at `lib/Debug/openxr_loader.lib`. For dll, it will be installed in `bin/Debug/openxr_loader.dll`.

I tested for both static and dynamic libraries.


rpavlik (approved):
Ah, that's a nice solution to that problem - I like it better than the libname suffix thing. Much less intrusive.

Thanks!

PR #207, opened 2020-09-08
Fix spelling of XrHandTrackerCreateInfoEXT.

aeikum:


yl-msft (approved):

rpavlik:
Ooh, good catch. How did my scripts not notice this? I must not have anything enforcing that the structextends struct exists...

rpavlik (approved):

aeikum:
I'm working on an OpenXR API bridge for Wine, based on [the winevulkan bridge generator script](https://source.winehq.org/git/wine.git/blob/master:/dlls/winevulkan/make_vulkan), since the XML spec formats are nearly identical. The generator script complained it couldn't find that structextends, which is how I noticed it.

rpavlik:
ah, cool! Looks like we just need the CLA to accept the merge. I've already made changes to our scripts to ensure this doesn't happen again.

BTW - I was looking at the script you linked. You may have an easier time with OpenXR - not only because there's no ICD pointer magic, but also in licensing sense: our headers are "Apache-2.0 OR MIT" (not just the registry XML) so you can use the unmodified headers. We also have a "reflection" header that contains a lot of metadata from the registry and which can be used for code generation directly in c/c++.

aeikum:
1000 words of legalese to change four bytes :+1: Go spray your lawyers with a squirt bottle.

Anyway should be good to go. Thanks for review, I'll be in touch if I come across anything else.

rpavlik (approved):

PR #208, opened 2020-09-15
Some formatting cleanup, remove trailing space characters and POSIX compliance

aaronfranke:
This PR does a few things related to file formatting:

* Remove trailing space characters. These are useless and just take up disk space. For most files I know off the top of my head that the change is fine, however, I wasn't sure about whether this is good for the `.adoc` and `.1` files, so I put the changes to those files in a 2nd commit with "(unsure)" at the end. If these are desired, I can squash them down.

* Ensure text files end in trailing newline characters. These are important for POSIX-compliance, since by POSIX definition a text file must end in a newline. Some Linux/Mac tools don't work properly without them present, including terminal apps and some C compilers. Also, GitHub shows a 🚫 warning for files that don't end in newlines.

* There were a few files that mistakenly contained BOMs, I removed them. BOMs are usually undesired.

I made these changes using a script. In case I need to rebase this PR, it is very easy to re-run the script I used. Additionally, if desired, I can add the script to the CI checks, so that these formatting checks are done automatically.


rpavlik (approved):
This is awesome, thanks! We'd love to have the script added - I had thought the gitattributes and the editorconfig files would take care of most of these things, but I will admit the cyclical disappearance-reappearance of the BOM in the xr.xml file is an annoying frustration...

I'd rather not touch the files in the license directory or the external files, but it seems strange they'd have trailing whitespace anyway, so I think it's probably fine. It might just happen again next time I update deps.

PR #209, opened 2020-09-19
Add file formatting checks to CI

aaronfranke:
Follow-up to #208. The formatting script runs on every text file except for those in `external/` and `src/external/`.

Fingers crossed that the CI checks will work, in previous implementations I had this script set up through GitHub Actions, but since this repo uses Azure Pipelines I wrote a config for that by copying the clang-format one. It doesn't seem that the new checks are running with this PR, that might change if it's merged but I'm not sure, I'm not familiar with Azure Pipelines.

### DIFF
--- a/.azure-pipelines/check_file_format.yml
+++ b/.azure-pipelines/check_file_format.yml
@@ -1,0 +1,24 @@
+# Copyright (c) 2020 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+jobs:
+- job: check_file_format
+  displayName: 'Check file formatting'
+  pool:
+    vmImage: 'ubuntu-20.04'
+  container: khronosgroup/docker-images:openxr-sdk
+  steps:
+    - script: sudo apt-get install -qq dos2unix recode
+      displayName: Install dependencies
+    - script: ./file_format.sh
+      displayName: File formatting checks (file_format.sh)
+
+    - script: git diff --patch --exit-code > file_format.patch
+      displayName: Save changes as diff
+    - script: echo "The following files need file formatting:"; sed -n -e "s/^diff.* b\///p" file_format.patch
+      condition: failed()
+    - task: PublishPipelineArtifact@1
+      displayName: Publish diff
+      condition: failed()
+      inputs:
+        path: $(System.DefaultWorkingDirectory)/file_format.patch
+        artifact: file_format_changes
--- a/file_format.sh
+++ b/file_format.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# This script ensures proper POSIX text file formatting and a few other things.
+# This is supplementary to clang-format.
+
+# We need dos2unix and recode.
+if [ ! -x "$(command -v dos2unix)" -o ! -x "$(command -v recode)" ]; then
+    printf "Install 'dos2unix' and 'recode' to use this script.\n"
+fi
+
+set -e -uo pipefail
+IFS=$'\n\t'
+
+# Loops through all text files tracked by Git.
+git grep -zIl '' |
+while IFS= read -rd '' f; do
+    # Exclude some files.
+    if [[ "$f" == "external"* ]]; then
+        continue
+    elif [[ "$f" == "src/external"* ]]; then
+        continue
+    fi
+    # Ensure that files are UTF-8 formatted.
+    recode UTF-8 "$f" 2> /dev/null
+    # Ensure that files have LF line endings and do not contain a BOM.
+    dos2unix "$f" 2> /dev/null
+    # Remove trailing space characters and ensures that files end
+    # with newline characters. -l option handles newlines conveniently.
+    perl -i -ple 's/\s*$//g' "$f"
+done
+
+# If no patch has been generated all is OK, clean up, and exit.
+if git diff --exit-code > patch.patch; then
+    printf "Files in this commit comply with the formatting rules.\n"
+    rm -f patch.patch
+    exit 0
+fi
+
+# A patch has been created, notify the user, clean up, and exit.
+printf "\n*** The following differences were found between the code "
+printf "and the formatting rules:\n\n"
+cat patch.patch
+printf "\n*** Aborting, please fix your commit(s) with 'git commit --amend' or 'git rebase -i <hash>'\n"
+rm -f patch.patch
+exit 1
--- a/.azure-pipelines/build_jobs.yml
+++ b/.azure-pipelines/build_jobs.yml
@@ -116,7 +116,7 @@ jobs:
       # Use the specified version of Python from the tool cache
       - task: UsePythonVersion@0
         inputs:
-          versionSpec: '3.9' 
+          versionSpec: '3.9'
       - task: PythonScript@0
         displayName: Move artifact contents
         inputs:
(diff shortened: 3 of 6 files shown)


file_format.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
```
rpavlik, line 1: Need copyright and license statement. If in doubt, use this one:

```
# Copyright (c) 2020 The Khronos Group Inc.
# SPDX-License-Identifier: Apache-2.0
```

.azure-pipelines/check_file_format.yml
```diff
@@ -0,0 +1,23 @@
+
```
rpavlik: Needs copyright/license: see other comment for sample.

file_format.sh
```diff
@@ -0,0 +1,49 @@
...
+    fi
+    # Ensure that files are UTF-8 formatted.
+    recode UTF-8 "$f" 2> /dev/null
+    # Ensure that files have LF line endings and do not contain a BOM.
+    dos2unix "$f" 2> /dev/null
+    # Remove trailing space characters and ensures that files end
+    # with newline characters. -l option handles newlines conveniently.
+    perl -i -ple 's/\s*$//g' "$f"
+    # Remove the character sequence "== true" if it has a leading space.
+    perl -i -pe 's/\x20== true//g' "$f"
```
rpavlik: This seems not necessary.

file_format.sh
```diff
@@ -0,0 +1,49 @@
...
+# We need dos2unix and recode.
+if [ ! -x "$(command -v dos2unix)" -o ! -x "$(command -v recode)" ]; then
+    printf "Install 'dos2unix' and 'recode' to use this script.\n"
+fi
+
+set -uo pipefail
+IFS=$'\n\t'
+
+# Loops through all text files tracked by Git.
+git grep -zIl '' |
```
rpavlik, line 17: is this basically supposed to do the same thing as `git ls-files`? Somehow it manages to break running from WSL - it's trying to access stuff using the Windows file path instead of the corresponding Linux one. Unfortunately it still gives the same error with git ls-files thru wsl.

```
ryanp@RYZENSHINE-WIN  E:\src-ssd\OpenXR-SDK-Source   format-ci ≢ +3 ~1 -0 !                                [09:56]
❯ wsl ./file_format.sh
fatal: cannot chdir to 'E:/src-ssd/OpenXR-SDK-Source': No such file or directory
fatal: cannot chdir to 'E:/src-ssd/OpenXR-SDK-Source': No such file or directory
Files in this commit comply with the formatting rules.
```

file_format.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+
+# This script ensures proper POSIX text file formatting and a few other things.
+# This is supplementary to clang-format.
+
+# We need dos2unix and recode.
+if [ ! -x "$(command -v dos2unix)" -o ! -x "$(command -v recode)" ]; then
+    printf "Install 'dos2unix' and 'recode' to use this script.\n"
```
rpavlik, line 10: why printf? and should we `exit 1` here? or just `exit 0` to avoid failing CI if I am missing the right stuff in the docker image?

changes/sdk/pr.206.gh.OpenXR-SDK-Source.md
```diff
@@ -1 +1 @@
-cmake: fix openxr_loader target export when installing both Release and Debug config on Windows.
\ No newline at end of file
+cmake: fix openxr_loader target export when installing both Release and Debug config on Windows.
```
rpavlik: please drop this change (when rebasing) to solve the conflict

file_format.sh
```diff
@@ -0,0 +1,49 @@
...
+fi
+
+set -uo pipefail
```
rpavlik: probably `set -e` in here too, since I'm getting:

```
ryan@RyZenShine-Win:/mnt/e/src-ssd/OpenXR-SDK-Source$ ./file_format.sh
fatal: cannot chdir to 'E:/src-ssd/OpenXR-SDK-Source': No such file or directory
fatal: cannot chdir to 'E:/src-ssd/OpenXR-SDK-Source': No such file or directory
Files in this commit comply with the formatting rules.
```

file_format.sh
```diff
@@ -0,0 +1,49 @@
...
+    # Ensure that files have LF line endings and do not contain a BOM.
+    dos2unix "$f" 2> /dev/null
+    # Remove trailing space characters and ensures that files end
+    # with newline characters. -l option handles newlines conveniently.
+    perl -i -ple 's/\s*$//g' "$f"
+    # Remove the character sequence "== true" if it has a leading space.
+    perl -i -pe 's/\x20== true//g' "$f"
+done
+
+git diff > patch.patch
```
rpavlik: I think `git diff --exit-code` can be your friend here - it's what we use in gitlab ci.

file_format.sh
```diff
@@ -0,0 +1,49 @@
...
+    rm -f patch.patch
+    exit 0
+fi
+
+# A patch has been created, notify the user, clean up, and exit.
+printf "\n*** The following differences were found between the code "
+printf "and the formatting rules:\n\n"
+cat patch.patch
+printf "\n*** Aborting, please fix your commit(s) with 'git commit --amend' or 'git rebase -i <hash>'\n"
+rm -f patch.patch
```
rpavlik, line 46: I think ideally we should archive the patch as an artifact, instead of deleting it, so it's easy to fix.

rpavlik (changes requested):
Thanks for this! I have a few comments, but I do like the general idea and I have used it before the upcoming release despite not merging it yet.

file_format.sh
aaronfranke: The difference between this code and `git ls-files` is that `git ls-files` includes binary files but the above code excludes them. We don't want to try converting a PNG file to UTF-8 for example.

I'm not sure what to do to make this work for WSL. The `git grep -zIl '' |` line works on Linux as-is, and `IFS=$'\n\t'` was added for compatibility with Mac and BSDs.

file_format.sh
aaronfranke: When modifying files locally, the easiest solution I could think of was to just have it delete the patch. If you want to archive this somewhere instead, that's a decent idea, but note that it's already printed to the terminal with `cat patch.patch`, so if you have access to the CI's terminal output then you're all set.

file_format.sh
aaronfranke: I don't remember why `printf` was used, it might be another case of BSD compatibility.

`exit 1` or `exit 0` are up to you. Without them, the script will warn you when these tools are missing but still run what it can. If there is an exit here, I think `exit 1` makes more sense.

file_format.sh
aaronfranke: Sure. `if git diff --exit-code > patch.patch; then` works fine.

file_format.sh
rpavlik: true, though I will confess it's probably challenging to copy and paste a mostly-whitespace or BOM patch from a terminal output log on a web page ;)

aaronfranke:
@rpavlik Requesting review and merge. I'm not sure what else I can do with this PR, I've responded to your feedback as well as I can. I think the best course of action is to merge it, and then you can make any further changes as needed.

rpavlik:
sorry for not merging this sooner - it looks good, I was just having some trouble getting it to run on our CI. I did run it before the last release, I think, despite not being able to merge. This one is on me now.

rpavlik (approved):

file_format.sh
rpavlik: ah, it was a git worktree config thing.

rpavlik (approved):

rpavlik:
Thanks - figured out my wsl problem and got it merged!

PR #213, opened 2020-10-11
Fix implicit API layer manifest example;

bjornbytes:
When using the implicit API layer example in the docs, the loader fails to load the layer properly, saying

```
Implicit layer /path/to/layer.json is missing "disable_environment"
```

After referring to other API layers, it looks like the key is supposed to be a string instead of an object.  Once this change is made to the manifest, the loader is able to load the layer.  This PR updates the docs to reflect that.

### DIFF
--- a/specification/loader/api_layer.adoc
+++ b/specification/loader/api_layer.adoc
@@ -415,12 +415,8 @@ only requirement is that the filename extension is ".json".
                "extension_version": "1"
            }
        ],
-       "enable_environment": {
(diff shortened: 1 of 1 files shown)


rpavlik:
This was discovered in the WG almost simultaneously with your submission, thanks! (I'll merge this but you'll see it change a little in the next release, just to match what the WG discussed and merged internally already)

rpavlik (approved):

PR #219, opened 2020-11-04
XrMatrix4x4f_CreateProjection: Explicitly define matrix values as floats

1runeberg:
Prevents potential division by zero, particularly lines 497, 503, 518 and 524


rpavlik:
Thanks!

PR #225, opened 2020-11-19
Include Win32 loader symbols in release artifact

brycehutchings:
Customers have requested access to the symbols that match the DLLs that we release (both through the .zip artifact and the NuGet package).


rpavlik (approved):
Looks reasonable to me - is it worth adding a changelog fragment? or maybe just adding to the one where I put more stuff in the release?

brycehutchings:
> Looks reasonable to me - is it worth adding a changelog fragment? or maybe just adding to the one where I put more stuff in the release?

Yup added!

rpavlik (approved):

rpavlik:
Have you tried using these PDBs? I tried debugging into it but didn't have any symbols:

image

brycehutchings:
> Have you tried using these PDBs? I tried debugging into it but didn't have any symbols:
> 
> image

That looks like it is using the loader statically linked in, since I don't see openxr_loader.dll in the stack. In that case you need to use the hello_xr.pdb symbol file.

rpavlik:
I literally just downloaded the binaries from the CI there. I think by the time it went boom, it was already out of the loader.

brycehutchings:
> I literally just downloaded the binaries from the CI there. I think by the time it went boom, it was already out of the loader.

If the loader was in the callstack and it was consumed as a DLL then the PDB would help, or if the developer "F11" steps in to a loader (openxr) function then the symbol will allow that.

PR #230, opened 2020-11-27
hello_xr: On Vulkan, explicitly add surface extensions for mirror window.

aeikum:
Some OpenXR providers may return these extensions from
xrGetVulkanInstanceExtensionsKHR, but they don't have to. Since they are
required for hello_xr to call vkCreateWin32SurfaceKHR, we should include
them explicitly.

### DIFF
--- a/changes/sdk/pr.230.gh.OpenXR-SDK-Source.md
+++ b/changes/sdk/pr.230.gh.OpenXR-SDK-Source.md
@@ -1,0 +1,1 @@
+hello_xr: On Vulkan, explicitly add surface extensions for mirror window.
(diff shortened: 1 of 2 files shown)


aeikum:
Notably, the Linux SteamVR OpenXR provider does *not* return any surface extensions. So we would definitely need this if mirror windows are ever supported on Linux, and it seems correct to do it on Windows, too.

rpavlik:
Thanks for this! I think I simultaneously did this internally as well, but I'll merge this and ensure you get credit too.

rpavlik (approved):

PR #239, opened 2021-02-01
Remove canonicalization in Windows loader

### LINKED ISSUE #214: Loader uses pathcch.lib which is only available on Windows 8 and later
This is an unintentional side-effect of #198. The loader should use something else that is available on Windows 7 too. Since symbolic links are not an important use case for the loader on Windows, perhaps std::filesystem::absolute can be used instead?

brycehutchings:
To fix the OpenXR loader on Windows 7, I am making the filesystem canonicalization function a no-op. Canonicalization was added for Linux support to follow a symlink, but symlinks are not used on Windows and there is no canonicalization function that can be used everywhere.

### DIFF
--- a/changes/sdk/pr.239.gh.OpenXR-SDK-Source.md
+++ b/changes/sdk/pr.239.gh.OpenXR-SDK-Source.md
@@ -1,0 +1,1 @@
+Loader: Fix loader failing to load on Windows 7 due to pathcch dependency.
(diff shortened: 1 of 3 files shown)


JoeLudwig (approved):

rpavlik (changes requested):
I think you added a submodule "DiligentEngine" by accident here. Also, good to mention the various issues in the front-matter of the changelog fragment.

brycehutchings:
> I think you added a submodule "DiligentEngine" by accident here. Also, good to mention the various issues in the front-matter of the changelog fragment.

Crap, my hobby project leaked in... good catch.

rpavlik (approved):

PR #242, opened 2021-03-04
Ensure we expose the OpenXR headers in the build interface as well as the install interface.

### LINKED ISSUE #195: OpenXR include directory included as PRIVATE, should be PUBLIC
https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/51cade1763267c750f53cefa9834644c7e7b37f3/src/loader/CMakeLists.txt#L91

If the openxr include directory were included as PUBLIC rather than PRIVATE for the openxr_loader target, it would simplify the process of importing that target using modern cmake's FetchContent. PUBLIC includes of a target are automatically included for any targets linking against that target.
Currently I have to manually include the directory instead.

Ref [this thread](https://gitlab.com/OpenMW/openmw/-/merge_requests/251#note_383337564) where we encountered the issue.

rpavlik:
Fixes #195 hopefully.  @madsbuvi can you confirm this is sufficient for you?

### DIFF
--- a/changes/sdk/pr.242.gh.OpenXR-SDK-Source.md
+++ b/changes/sdk/pr.242.gh.OpenXR-SDK-Source.md
@@ -1,0 +1,5 @@
+---
+- issue.195.gh.OpenXR-SDK-Source
+- issue.1409.gl
+---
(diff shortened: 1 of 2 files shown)


madsbuvi:
This fixes the issue, yes. Thank you

rpavlik:
great, thanks for following up!

PR #243, opened 2021-03-08
Registry XML fixes

### LINKED ISSUE #80: Registry XML contains unescaped '>' character
Presumably this should have been escaped as `&gt;` like other such symbols in the same file.

https://github.com/KhronosGroup/OpenXR-Docs/blob/master/specification/registry/xr.xml#L124

I've also noticed that this XML contains an UTF-8 byte-order-mark, which the vulkan xml doesn't, though I'm not sure if this is expected to be consistent between these two or not.

s-ol:
Close KhronosGroup/OpenXR-Docs#80 and KhronosGroup/OpenXR-Docs#81.

### DIFF
--- a/specification/registry/xr.xml
+++ b/specification/registry/xr.xml
@@ -1345,8 +1345,8 @@ maintained in the default branch of the Khronos OpenXR GitHub project.
         <type category="struct" name="XrControllerModelNodePropertiesMSFT">
             <member values="XR_TYPE_CONTROLLER_MODEL_NODE_PROPERTIES_MSFT"><type>XrStructureType</type> <name>type</name></member>
             <member><type>void</type>* <name>next</name></member>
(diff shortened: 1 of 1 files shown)


rpavlik:
Thanks for this! I'll have to transfer it to the OpenXR-Docs repo to merge it, but I'm glad to do so! In the future, you may also choose to write your own changelog fragment for any changes you make.

rpavlik:
Ah, right - that repo has a different CLA because it's specification content rather than mainly software. Please head over to https://github.com/KhronosGroup/OpenXR-Docs/pull/85 and hit the CLA assistant link there when you get the chance, so I can merge your contribution.

s-ol:
Thanks! Yeah I got a bit confused and realized too late that the docs repo, where I openes the original issues, was actually the right place to send this to. Thanks for moving the changes over, the CLA is now signed as well.

PR #245, opened 2021-03-11
Update BUILDING.md

muchcharles:
VS2019 build instructions, including some gotchas


rpavlik:
Looks nice, thanks! You can also just open the project with Visual Studio "open folder" in 2019. You shouldn't actually need to use the visual studio command prompt or care about where msbuild is on your path, I clearly haven't reviewed these instructions in a long time. (You might need to use `cmake --build . --config RelWithDebInfo` instead of directly calling msbuild if not using the developer command prompt, but if you use a 'visual studio' named generator with CMake, you don't need anything in the environment to point to the compiler).

rpavlik (approved):

PR #246, opened 2021-03-11
Fix warning: directive in .EXP differs from output filename

brycehutchings:
From what I know, the filename specified after LIBRARY in a .def file is optional and can be removed. The name conflicts with the debug build since the "d" suffix is included with the filename and results in the following warning:

openxr_loaderd.exp : warning LNK4070: /OUT:openxr_loader.dll directive in .EXP differs from output filename '(...)\OpenXR-SDK-Source\build\src\loader\Debug\openxr_loaderd.dll'; ignoring directive


bradgrantham-lunarg (approved):

rpavlik:
Ohhh, weird. Good catch!

rpavlik (approved):

PR #247, opened 2021-03-20
Add header include

iesser:
This project compiles by chance because the header "openxr_reflection.h" was included before the #include "common.h".
It would be better to explicitly include this file here, especially because it is needed to get the macro "to_string" working.


rpavlik (approved):

rpavlik:
Looks great, thanks! I've added a changelog fragment for it.

PR #251, opened 2021-04-02
Disable extension prototypes

### LINKED ISSUE #174: Methods defined in openxr_platform.h  such as XrGraphicsRequirementsOpenGLESKHR have no impemention in openxr_loader.so
Hi  rpavlik:
I find methods defined in openxr_platform.h  such as XrGraphicsRequirementsOpenGLESKHR have no implementationin openxr_loader.so. when i call these method,i encount a Compile  error msg: undefined reference to 'xrGetOpenGLESGraphicsRequirementsKHR'.
should i implement these methods in loader_core.cpp or  somewhere.
 thanks very much

rpavlik:
Since the loader doesn't actually export extension functions, it can be confusing that the headers expose the extension prototypes by default. This change makes it so that you have to define `XR_EXTENSION_PROTOTYPES` (and not define `XR_NO_PROTOTYPES`) in order to get the prototypes for extension functions. (The function pointer types are not guarded this way, so those continue to work as expected.)

Fixes #174 among others


specification/scripts/cgenerator.py
```diff
@@ -213,15 +237,20 @@ def endFeature(self):
...
                                   self.genOpts.protectProtoStr, file=self.outFile)
+                        if self.genOpts.protectExtensionProto and not is_core:
+                            write(self.genOpts.protectExtensionProto,
+                                  self.genOpts.protectExtensionProtoStr, file=self.outFile)
                         write('\n'.join(self.sections['command']), end='', file=self.outFile)
-                        if self.genOpts.protectProto:
-                            write('#endif', file=self.outFile)
-                        else:
-                            self.newline()
+                        if not is_core:
```
brycehutchings: does this condition need to have "self.genOpts.protectExtensionProto and" to balance the condition on line 240?

brycehutchings: Ah writeProtectEnd will short circuit early if the string is null so it is ok that the `if` condition is different

brycehutchings: It might be cleaner if we also had a writeProtectBegin that behaved similarly

brycehutchings (approved):

specification/scripts/cgenerator.py
rpavlik: yeah I wasn't sure what to do here. It looks really messy to me, but I'm not sure my cleanup made it less messy. We do have a begin_protect and end_protect when we're using Jinja for generating stuff. Should I just undo the cleanup and leave it verbose?

rpavlik: OK, I revised this substantially to be more like the old code, hopefully it's more readable. Let me know your thoughts.

rpavlik:
Updated, ready to re-review. Generated headers look like this now:

image

brycehutchings (approved):

PR #252, opened 2021-04-08
Add locking to protect against multithreaded conformance test usage.

brycehutchings:
The "multithreaded" conformance test is intended to test the runtime but it has exposed two race conditions in the loader:
1. If debug utils extension is enabled, and instance-specific logger will be added and removed when an XrInstance is created. However, if another thread is trying to log while the list of loggers changes then it may crash. So I added a read/write lock in the logger class to protect this.
2. If an XrInstance is being destroyed while xrEnumerateInstanceProperties is called, xrDestroyInstance may unload the runtime while xrEnumerateInstanceProperties is calling into the runtime to get instance properties. I consolidated the two existing locks into a "global loader lock" and expanded its use to a few more areas to solve this. I didn't want to introduce a third lock because then you can easily get into accidental dead lock issues if the locks are acquired in the wrong order, so a single lock is safer and the functions that use the lock don't require special concerns around concurrency anyway.

### DIFF
--- a/src/loader/loader_core.cpp
+++ b/src/loader/loader_core.cpp
@@ -32,16 +32,12 @@
 #include <utility>
 #include <vector>
 
-// Global lock to prevent reading JSON manifest files at the same time.
-std::mutex &GetLoaderJsonMutex() {
-    static std::mutex loader_json_mutex;
-    return loader_json_mutex;
-}
-
-// Global lock to prevent simultaneous instance creation/destruction
-std::mutex &GetInstanceCreateDestroyMutex() {
-    static std::mutex instance_create_destroy_mutex;
-    return instance_create_destroy_mutex;
+// Global loader lock to:
+//   1. Ensure ActiveLoaderInstance get and set operations are done atomically.
+//   2. Ensure RuntimeInterface isn't used to unload the runtime while the runtime is in use.
+std::mutex &GetGlobalLoaderMutex() {
+    static std::mutex loader_mutex;
+    return loader_mutex;
 }
 
 // Terminal functions needed by xrCreateInstance.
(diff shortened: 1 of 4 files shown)


rpavlik:
@ChristophHaag can you help out with a review on this?

src/loader/loader_core.cpp
```diff
@@ -248,7 +245,6 @@ XRAPI_ATTR XrResult XRAPI_CALL LoaderXrCreateInstance(const XrInstanceCreateInfo
 
     // Make sure only one thread is attempting to read the JSON files and use the instance.
```
ChristophHaag, line 246: Is this comment now stale?
If the reason for the block below was the scope for locking, it doesn't need to be a block anymore now.

src/loader/loader_core.cpp
```diff
@@ -117,8 +113,8 @@ XRAPI_ATTR XrResult XRAPI_CALL LoaderXrEnumerateInstanceExtensionProperties(cons
     XrResult result;
 
     {
-        // Make sure only one thread is attempting to read the JSON files at a time.
-        std::unique_lock<std::mutex> json_lock(GetLoaderJsonMutex());
+        // Make sure the runtime isn't unloaded while this call is in progress.
```
ChristophHaag, line 116: minor nitpick: Maybe the original comment could be preserved since that lock does both of these things now

ChristophHaag (commented):
I haven't checked if those locks really lock everything that needs to be locked but I tried it with the gcc thread sanitizer and this PR fixes almost all of the issues it reports.

The code looks good safe for the two minor possible improvements I noted.

src/loader/loader_core.cpp
rpavlik: Neither Bryce nor I could figure out *why* it mattered that only one thread would read the json at a time, though. We suspect it was left over from when the loader was more ambitious and trying to do multiple instances and comparing runtimes

ChristophHaag: Sounds like the only real reason would be if the json parser was not threadsafe/had global state. I suppose that is not the case?`

rpavlik (approved):

PR #253, opened 2021-04-09
BUILDING.md: Typo fix (tailed -> tailored)

robbiesri:
Just stumbled on this while getting set up with the repo!

Had to reformat to keep it within 80 columns, though the formatting isn't consistent.


rpavlik:
great catch, thanks!

PR #274, opened 2021-10-12
Android file ifdef guard in loader

rpavlik:
Adding this makes it possible to continue just globbing all the files in OpenXR-SDK to build, even on non-Android.

### DIFF
--- a/changes/sdk/pr.274.gh.OpenXR-SDK-Source.md
+++ b/changes/sdk/pr.274.gh.OpenXR-SDK-Source.md
@@ -1,0 +1,1 @@
+- Loader: Add `ifdef` guards around contents of Android-specific files so all platforms may still glob all source files in OpenXR-SDK to build the loader with a custom build system.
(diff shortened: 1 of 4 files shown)


brycehutchings:
Related to this change,

> "Where the elementCapacityInput is non-zero but less than required, the function sets elementCountOutput to the required capacity, and must return XR_ERROR_SIZE_INSUFFICIENT. The data in elements is undefined."

can be simplified to just

> "Where the elementCapacityInput is less than required, the function must return XR_ERROR_SIZE_INSUFFICIENT. The data in elements is undefined.

An earlier bullet already states that `elementCountOutput` is always written

rpavlik:
I think you meant to put this on https://github.com/KhronosGroup/OpenXR-Docs/pull/104

> 
> 
> Related to this change,
> 
> > "Where the elementCapacityInput is non-zero but less than required, the function sets elementCountOutput to the required capacity, and must return XR_ERROR_SIZE_INSUFFICIENT. The data in elements is undefined."
> 
> can be simplified to just
> 
> > "Where the elementCapacityInput is less than required, the function must return XR_ERROR_SIZE_INSUFFICIENT. The data in elements is undefined.
> 
> An earlier bullet already states that `elementCountOutput` is always written

Except that it doesn't say what is written: if it's insufficient, the required capacity is written. If it's sufficient, then the actual number populated is written. I agree we can simplify this, but not quite this much :)

PR #277, opened 2021-10-29
Defer Vulkan CPU sync until the next frame begins

Ralith:
Reduces the impact of the pipeline stall by moving it after presentation, and provides a simple test for whether a runtime properly synchronizes presentation with application work.


rpavlik:
Thanks, merging! (had a chance to test it with Monado finally)

rpavlik (approved):

PR #278, opened 2021-10-29
[loader]doc: Use nativeLibraryDir to replace getNativeLibraryDir()

utzcoz:
`ApplicationInfo` doesn't have method called `getNativeLibraryDir()`, and it
only has `nativeLibraryDir` field, used by OpenXR loader source code.
See https://developer.android.com/reference/android/content/pm/ApplicationInfo#nativeLibraryDir and https://gitlab.freedesktop.org/monado/utilities/openxr-android-broker/-/blob/main/utils/src/main/java/org/khronos/openxr/runtime_broker/utils/RuntimeData.java#L257-261.

### DIFF
--- a/specification/loader/runtime.adoc
+++ b/specification/loader/runtime.adoc
@@ -440,7 +440,7 @@ If more than one is found, user preferences are used to identify the
 "active" runtime.
 
 The path containing the dynamic library is computed from
-`ApplicationInfo.getNativeLibraryDir()` and the specified ABI, and the
+`ApplicationInfo.nativeLibraryDir` and the specified ABI, and the
 filename is returned using the filename found in the OpenXR metadata value.
 The "hasFunctions" column is dynamically generated based on the presence of
 any function metadata entries.


utzcoz:
Looks like the failed checking is caused by latest merged PR https://github.com/KhronosGroup/OpenXR-SDK-Source/runs/4048125162.

rpavlik:
Ah good catch, my jni wrapper script appends an "add" on property names to make a getter.

CI build actually broke earlier, I think.

utzcoz:
Hi @rpavlik , I rebased it on latest master branch with your ci-fix-up commits.

rpavlik:
So I realized this actually needs to be proposed in the OpenXR-Docs repo instead of here: https://github.com/KhronosGroup/OpenXR-Docs 

(for CLA reasons)

If you could transplant your patch to that repo, I'd greatly appreciate it.

utzcoz:
> So I realized this actually needs to be proposed in the OpenXR-Docs repo instead of here: https://github.com/KhronosGroup/OpenXR-Docs
> 
> (for CLA reasons)
> 
> If you could transplant your patch to that repo, I'd greatly appreciate it.

No problem. I will create a new issue about it at OpenXR-Docs today.

utzcoz:
Hi @rpavlik , as discussion at https://github.com/KhronosGroup/OpenXR-Docs/issues/108#issuecomment-962117029, could you help to review this PR again? Thanks.

rpavlik:
Yep I'll merge it probably today once I make a changelog fragment for it

rpavlik:
Thanks!

utzcoz:
> Thanks!

Thanks for your reviewing.

PR #280, opened 2021-11-04
Update jni wrappers, jnipp

### LINKED ISSUE #275: `android-jni-wrappers` error "use of undeclared identifier 'assert'"
The current master did not compile for Android cross-compiling as there is a missing `#include <cassert>` in the `android-jni-wrappers` code.

The full compiler error is as follows but occurs in multiple places:

```
C:\Projects\HelloXR\out\deps_src\openxr\src\external\android-jni-wrappers\wrap\ObjectWrapperBase.h(262,5): error G30564CF7: use of undeclared identifier 'assert'
      assert(!obj.isNull());
      ^
```

I'm guessing this is 'external' code by the path so maybe needs propagating up stream?

rpavlik:
Fixes #275 . Also fixes an issue where an app could cause a crash by detaching from the thread before shutting down OpenXR.

### DIFF
--- a/changes/sdk/pr.280.gh.OpenXR-SDK-Source.1..md
+++ b/changes/sdk/pr.280.gh.OpenXR-SDK-Source.1..md
@@ -1,0 +1,1 @@
+- Update jnipp to fix crash on Android if app detatches thread from JVM (e.g. on shutdown).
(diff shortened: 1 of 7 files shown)


rpavlik:
Reviewed internally by WG members.

PR #284, opened 2021-12-07
Loader fixes

rbessems:
Fixed loading of runtime on AOSP without a broker.

### DIFF
--- a/src/loader/manifest_file.cpp
+++ b/src/loader/manifest_file.cpp
@@ -638,7 +638,7 @@ XrResult RuntimeManifestFile::FindManifestFiles(std::vector<std::unique_ptr<Runt
                 "", "RuntimeManifestFile::FindManifestFiles - failed to determine active runtime file path for this environment");
             return XR_ERROR_RUNTIME_UNAVAILABLE;
         } else {
-          result = XR_SUCCESS;
+            result = XR_SUCCESS;
         }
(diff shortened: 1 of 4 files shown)


src/loader/manifest_file.cpp
```diff
@@ -637,6 +637,8 @@ XrResult RuntimeManifestFile::FindManifestFiles(std::vector<std::unique_ptr<Runt
             LoaderLogger::LogErrorMessage(
                 "", "RuntimeManifestFile::FindManifestFiles - failed to determine active runtime file path for this environment");
             return XR_ERROR_RUNTIME_UNAVAILABLE;
+        } else {
+            result = XR_SUCCESS;
```
rpavlik: Can we move this out of "else" and just put it at the same level as the debug message below? The else took me a bit to realize it was just setting a value, but not actually returning on its own unlike the "if" half. (If I were a static analyzer I'd say something about no need to use "else" after "return".)

rbessems: Done!

rpavlik:
Looks great, thanks! Would you mind putting a brief changelog fragment in changes/sdk/pr.284.gh.OpenXR-SDK-Source.md? See the readme in that directory for instructions.

rpavlik (approved):

rpavlik (approved):

PR #285, opened 2021-12-07
Jnipp fixes

rbessems:
Fixing JNI null de-reference crash.


rpavlik:
Thanks for this! I actually had just gotten essentially this patch submitted to Monado, but based on and including a few other small fixes https://gitlab.freedesktop.org/monado/monado/-/merge_requests/1032  I'll close this PR but will reference it in the changelog fragment for the one I submitted.

PR #319, opened 2022-06-28
Add vendor ID

fredemmott:
Software that will use this vendor id: https://github.com/fredemmott/OpenKneeboard/ - currently uses XR_APILAYER_NOVENDOR_OpenKneeboard.

Spec notes that vendor IDs should be reserved, and can be used for API layers - and, if you have a vendor ID, it *should* be used for all your API layers. However, it's unclear if it's fine to publish an API layer without an assigned vendor ID:

> This usually takes the form of:
>
> An OpenXR API layer prefix string: "XR_APILAYER_"
>
> The name of the vendor developing the API layer: e.g. "LUNARG_"
>
> A short descriptive name of the API layer: e.g. "test"

If it is intended that api layers 'should' or 'must' register vendors, please approve this pull request, so I can use it in my software.

If this is not intended, perhaps the documentation should encourage the use of `NOVENDOR` or similar? This is used by several other projects, e.g. OpenXR toolkit

### DIFF
--- a/changes/registry/pr319.gh.OpenXR-SDK-Source.md
+++ b/changes/registry/pr319.gh.OpenXR-SDK-Source.md
@@ -1,0 +1,1 @@
+Register author ID for Fred Emmott.
--- a/specification/registry/xr.xml
+++ b/specification/registry/xr.xml
@@ -34,6 +34,7 @@ maintained in the default branch of the Khronos OpenXR GitHub project.
         <tag name="EXT"        author="Multivendor"                 contact="Ryan Pavlik @rpavlik"/>
         <tag name="EXTX"       author="Multivendor experimental"    contact="Brad Grantham @bradgrantham-lunarg"/>
         <tag name="FB"         author="Facebook"                    contact="Cass Everitt @casseveritt, Jonathan Wright @Nelno"/>
+        <tag name="FREDEMMOTT" author="Frederick Emmott"            contact="Frederick Emmott @fredemmott"/>
         <tag name="GOOGLE"     author="Google"                      contact="Kaye Mason @chaleur"/>
         <tag name="HTC"        author="HTC"                         contact="Chris Kuo @ggkuo, Kyle Chen @kylechen76"/>
         <tag name="HTCX"       author="HTC"                         contact="Chris Kuo @ggkuo, Kyle Chen @kylechen76"/>


rpavlik:
Very cool, thanks for submitting this! Happy to see more creative OpenXR software!

I think for API layers, if you aren't exposing a custom extension, you can just use a reverse-DNS-name style thing (io.github.fredemmott -> io_github_fredemmot)  without registration:

> Details on how to register an author ID are provided below. API layer authors not wishing to register an author ID with Khronos can instead use a fully-qualified domain name (FQDN) as the ID. The FQDN should be a domain name owned by the author. FQDNs cannot be used for extensions, only for API layers.

https://www.khronos.org/registry/OpenXR/specs/1.0/styleguide.html#extensions-naming-conventions-name-strings

(maybe it's not reverse order?)

rpavlik:
oh, it is reverse order, it's just written further down.

> Because API layers need not be registered with Khronos, an alternative mechanism is needed to allow creating unique API layer names without registering an author ID. API layer authors that prefer not to register an author ID can instead use a fully-qualified domain name (FQDN) in reverse-order as an author ID, replacing . (period) with _ (underscore) characters. The restriction that API layer names must be valid C identifiers means that some FQDNs cannot be used as part of API layer names.

This could probably use some cleanup so it's easier to find.

fredemmott:
Thanks, that works - though 'not wishing to register' implies this is a matter of author preference rather than necessarily the recommended approach?

rpavlik:
I'll ask the WG what they think, and ping the OpenXR Toolkit layer author too since I know them.

rpavlik:
The group agrees that registering an author ID is preferred so we'll leave the text fundamentally alone.

Can you add a changelog fragment file in changes/registry/pr.319.gh.OpenXR-SDK-Source.md with something like "Register author ID for ..."

fredemmott:
Thanks, done :) I've just entered that as a line by itself, basing it on a few entries I saw in the git history. Let me know if you need the header block also containing `pr.319.gh.OpenXR-SDK-Source` as the README suggests

rpavlik:
Oh, the readme shouldn't suggest that. You only need the header thing if you want to link more than one item to it. Looks good, thanks!

fredemmott:
Hi :) this still has 'waiting for reporter' and 'needs changelog fragment' - I think that's done, do you need anything else from me?

rpavlik:
Oh, no it just needs the target branch changed to staging, which I apparently can't do from mobile and the automation we have isn't working right. If you can do that, I can merge it very soon, otherwise I'll get to it when I'm back in the office here soon.

fredemmott:
Thanks! it doesn't look like a staging branch exists at the moment, so I'm not able to do it:

image

rpavlik:
Oh, rats. That's because this is the wrong repo, sorry. OpenXR-Docs is the place for modifying the registry. The only reason it matters is I think they have a different CLA.

fredemmott:
Thanks - closing in favor of https://github.com/KhronosGroup/OpenXR-Docs/pull/131

PR #324, opened 2022-08-07
disable OPENGL on macOS

kcking:
OpenGL is causing lots of compilation errors on macOS, and it is deprecated anyways, so it seems most pragmatic to me to just disable it.

My original PR built successfully with `release-1.0.22`, but once rebased on top of main, required this fix to compile successfully.

Let me know if this is desirable and I can tack on a change fragment as well. Thanks!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -50,6 +50,8 @@ elseif(ANDROID)
         set(VULKAN_INCOMPATIBLE TRUE)
         message(STATUS "Vulkan disabled due to incompatibility: need to target at least API 24")
     endif()
+elseif(APPLE)
+    set(OPENGL_INCOMPATIBLE TRUE)
 endif()
 
 if(NOT OPENGL_INCOMPATIBLE)


rpavlik:
So what graphics API are you using if you disabled OpenGL? Alternately is there a way to work out the fix to make it build again? Trying to think of what changed between 1.0.22 and 1.0.24 that might affect it, I can't say there was much software change I recall, would have to look at the changelog and diff

kcking:
Oh I have been using Vulkan -> Molten, I wasn't aware whether OpenGL on macOS was considered a supported platform combination. The reason I never tried it was because the libraries I'm using use `wgpu` which doesn't support OpenGL

rpavlik:
Ah ok. I'd still be interested in getting OpenGL build fixed.

kcking:
Sorry for the noise, the build issues appear to have been a false alarm

rpavlik:
oh excellent! glad to hear it.

PR #332, opened 2022-09-19
loader: Report STL usage as "none" due to our trimmed exports.

rpavlik:


johnkearney:
Thanks Ryan! Looks good.

rpavlik:
Prefab maintainer confirmed that "none" is the value we should use. I just need to add a change log fragment

rpavlik:
Added changelog fragment, no other change in the latest push.

PR #336, opened 2022-09-29
Debug message when secure environment variables are ignored

brycehutchings:
The loader will ignore environment variables that determine which code to load when running from an elevated context (Windows) or secure execution is required (Linux). This has caught many people off guard and so this is the first step to try and make this behavior more visible. When this happens a message will be written into the debugger and to stderr like so (see first line):

image

### DIFF
--- a/changes/sdk/pr.336.gh.OpenXR-SDK-Source.md
+++ b/changes/sdk/pr.336.gh.OpenXR-SDK-Source.md
@@ -1,0 +1,1 @@
+Loader and layers: In debug builds, log when non-empty environment variables are being ignored due to executing with elevated privilege.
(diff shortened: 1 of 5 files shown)


rpavlik:
So the azure Dev ops failures are all the Linux builds, plus some kind of schema failure with the license metadata for the nupkg?! Not sure what happened there, maybe some schema change came in with a system update

rpavlik:
Oh here's the Linux error 

```
./src/common/platform_utils.hpp: In function ‘std::__cxx11::string PlatformUtilsGetSecureEnv(const char*)’:
../src/common/platform_utils.hpp:86:18: error: request for member ‘empty’ in ‘str’, which is of non-class type ‘char*’
         if (!str.empty()) {
```

rpavlik:
OK, I fixed the linux build, you should be able to rebase on main soon here and that will fix the remaining CI issue.

rpavlik (changes requested):
This looks good to me. Please add a changelog fragment.

rpavlik (approved):

rpavlik:
I added the changelog fragment. WG has approved change for merge.

PR #347, opened 2022-11-16
cmake: Use standard `CMAKE_INSTALL_INCLUDEDIR`

KiruyaMomochi:
Current code uses hard-coded `include` or `CMAKE_INSTALL_INCDIR` to specify included directories. It makes the installed target not work with [Nix](https://nixos.org/), which tries to split headers into a different package. 

https://github.com/NixOS/nixpkgs/blob/7cb6229334cecda7592a6df2a23a2d19549044f1/pkgs/build-support/setup-hooks/multiple-outputs.sh#L150-L165

`CMAKE_INSTALL_INCLUDEDIR` is CMake’s [recommended](https://cmake.org/cmake/help/latest/command/install.html#installing-targets) standard for setting header location. It allows package maintainers like Nix to control the install destination by setting appropriate cache variables.

https://github.com/NixOS/nixpkgs/blob/7cb6229334cecda7592a6df2a23a2d19549044f1/pkgs/development/tools/build-managers/cmake/setup-hook.sh#L91-L103


rpavlik:
Good catch, I'm not sure why we weren't using that right, I definitely know and have used CMAKE_INSTALL_INCLUDEDIR. We do have to override it if we're packaging for an Android AAR file.

rpavlik (approved):
Was slightly nervous this would break the AAR/prefab generation for Android, but I tested locally and it looks fine. Thanks!

rpavlik:
Would you mind adding a changelog fragment in the changes/sdk directory named `pr.347.gh.OpenXR-SDK-Source.md` ?

rpavlik:
Awesome, thanks!

PR #362, opened 2022-12-07
 MinGW build fix

brainrom:
`.def` and `.pdb` are MSVC-specific resources. MinGW can correctly build the OpenXR loader, but only with these changes.


rpavlik:
The one thing I'm not sure about, is that I am pretty sure that LLVM/Clang on Windows can make PDB's. I was hoping CMake would have a variable or property that says "this toolchain makes pdb" but I haven't found one.  This is an improvement but I wonder if we can flip the condition, only exclude for mingw?

brainrom:
Hmm. Okay, then `if(WIN32 AND NOT MINGW AND DYNAMIC_LOADER)`?

brainrom:
Changed the condition.

rpavlik (approved):

rpavlik:
Thanks very much!

rpavlik:
We received #387 to correct the exported symbols, which effectively undoes this. Can you work with that submitter to figure out the correct answer here that will work for both of you?

PR #381, opened 2023-03-07
Add VS 2022 version code to BUILDING.md

utzcoz:
The VS 2022 version code 17.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -51,6 +51,7 @@ number. The following table is provided to help you:
 | Visual Studio 2015   |       14       |
 | Visual Studio 2017   |       15       |
 | Visual Studio 2019   |       16       |
+| Visual Studio 2022   |       17       |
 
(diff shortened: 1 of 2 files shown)


utzcoz:
Looks like `cmake -G "Visual Studio [Version Number] Win64" ../..` is not correct because latest CMake wants something like `Visual Studio 17 2022`.

utzcoz:
Hi @rpavlik , could you help review this PR? Thanks.

rpavlik:
Yep, thanks! I keep meaning to remove this section and just point to the Cmake user instructions but until I get around to it this will help 😁

utzcoz:
@rpavlik I also want to replace it with GitHub Actions' CMake building steps. It might needs extra steps to setup MSVC environment. At least, I can build HelloXR and Loader with CMake + MSVC environment locally. I decide to send a PR for it this weekend if need.

PR #383, opened 2023-03-11
build: Move CMakePresets.json to the top directory

utzcoz:
If developers want to build HelloXR on Windows platform with CMake, they might want to copy `CMakePresets.json` to the top directory for convenience. This copy command might cause dirty git diff. This CL just moves it to the top directory to remove this extra copying command for developers.


utzcoz:
Hi @rpavlik , could you help to review this PR? Thanks.

rpavlik:
So having it there actually makes it harder to build with vscode, I regret having it there for Monado. (You have to launch vscode from a developer command prompt. Also makes it impossible to use vscode remote ssh into a Windows machine.) The only reason it's in the repo is it makes the CI easier. If it's confusing I would rather remove it than publicize it more, the preset thing isn't ready for widespread use yet, at least with vscode, and visual studio isn't much better with it.

(I really should file issues with the vscode extension, etc so they can actually fix it, since the idea is nice and it works well on Linux usually.)

utzcoz:
@rpavlik I like the idea of `CMakePresets.json`, and looks like it can help us to build HelloXR on Windows with normal Powershell terminal instead of "Developer Powershell Prompt for VS 2022". 

> (I really should file issues with the vscode extension, etc so they can actually fix it, since the idea is nice and it works well on Linux usually.)

I think `CMakePresets.json` is the official feature of CMake, what about filing an issue to let vscode to support it?

rpavlik:
Yeah, it is "supported", but it turns off all the kit-finding the vscode extension normally does, which is why you have to launch from a dev command prompt. I'd be quite happy if we could get this resolved :)

utzcoz:
@rpavlik Another choice is to remove `CMakePreset.json` and use Monado's ps1 script to build this repository on Windows? Maybe I can give a try.

utzcoz:
Looks like [CLion detects `CMakePresets.json` fromt the top directory default](https://www.jetbrains.com/help/clion/cmake-presets.html#detect). So I reopen this PR.

rpavlik:
Hoping for a resolution to https://github.com/microsoft/vscode-cmake-tools/issues/3079 first, since the current status is really painful for me to do Windows dev with a presets file.

PR #387, opened 2023-03-25
cmake: Fix exporting APIs from openxr-loader in mingw

Biswa96:
cmake: Fix exporting APIs from openxr-loader in mingw

    This adds openxr-loader.def in source list of openxr-loader.dll for
    all toolchain in Windows platform and fixes linking with test programs.
    Otherwise, all required APIs are not exported from the shared library
    and compiling test programs shows following errors

    ld.exe: list_json.cpp.obj(.text+0x686): undefined reference to `xrDestroyInstance'
    ld.exe: list_json.cpp.obj(.text+0x96f): undefined reference to `xrCreateInstance'

### DIFF
--- a/src/loader/CMakeLists.txt
+++ b/src/loader/CMakeLists.txt
@@ -41,7 +41,7 @@ run_xr_xml_generate(loader_source_generator.py xr_generated_loader.cpp)
 if(DYNAMIC_LOADER)
     add_definitions(-DXRAPI_DLL_EXPORT)
     set(LIBRARY_TYPE SHARED)
-    if(MSVC)
+    if(WIN32)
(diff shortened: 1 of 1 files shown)


rpavlik:
The opposite of this was recently merged in #362, so I am not quite sure what to do here. It would be ideal if we could get a GitHub Actions CI job added to actually test these cases, since I do not regularly use mingw when building for/on Windows.

Biswa96:
> It would be ideal if we could get a GitHub Actions CI job added to actually test these cases

Agree. I can try to add the CI job.

Biswa96:
Should I create a separate pull request for the mingw GitHub Actions CI job?

rpavlik:
Yes, please.

PR #395, opened 2023-04-05
hello_xr: Make Vulkan's mirror window optional

### LINKED ISSUE #392: HelloXR Vulkan version enables USE_MIRROR_WINDOW default, and it is annoying when testing it on Windows
See https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/main/src/tests/hello_xr/graphicsplugin_vulkan.cpp#L21.

From the comment, `USE_MIRROR_WINDOW` shows a dummy window when running Vulkan version of HelloXR on Windows platform, and this window is used for RenderDoc. As a normal tester, I don't use RenderDoc, and I just want the Runtime to display a native window to show the rendered content. What about disabling it default? If other folks want to run it with RenderDoc, they can build it with patch to enable `USE_MIRROR_WINDOW` macro.

utzcoz:
I decide to disable it default.

Fixes: https://github.com/KhronosGroup/OpenXR-SDK-Source/issues/392.

### DIFF
--- a/changes/sdk/pr.395.gh.OpenXR-SDK-Source.md
+++ b/changes/sdk/pr.395.gh.OpenXR-SDK-Source.md
@@ -1,0 +1,1 @@
+hello_xr: Make Vulkan's mirror window optional
--- a/src/tests/hello_xr/graphicsplugin_vulkan.cpp
+++ b/src/tests/hello_xr/graphicsplugin_vulkan.cpp
@@ -16,11 +16,6 @@
 #include <shaderc/shaderc.hpp>
 #endif
 
-#if defined(VK_USE_PLATFORM_WIN32_KHR)
-// Define USE_MIRROR_WINDOW to open a otherwise-unused window for e.g. RenderDoc
-#define USE_MIRROR_WINDOW
-#endif
-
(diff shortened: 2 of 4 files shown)


rpavlik:
The WG would prefer having it as a command-line argument, suspecting that the next person who wants to use it will submit a PR re-enabling it :rofl: 

Making it a configuration option in CMake might be a useful halfway point I forgot to raise during the call.

utzcoz:
> Making it a configuration option in CMake might be a useful halfway point I forgot to raise during the call.

Do I need to support it at this PR or next PR. It is all fine to me.

rpavlik:
I think they didn't want to merge this one, so probably either keep this patch locally or try a Cmake or runtime option.

utzcoz:
> I think they didn't want to merge this one, so probably either keep this patch locally or try a Cmake or runtime option.

No problem. I will try to improve this PR with cmake or runtime option.

utzcoz:
Hi @rpavlik , I updated the patch to make mirror window optional with new command argument: --mirrorwindow|-mw. Please help review new patch to check what extra things that I need to do to meet requirements. Thanks.

PR #396, opened 2023-04-05
ci: Add mingw build in GitHub Actions

Biswa96:


Biswa96:
Should I fix the mingw build issue in this pull request?

rpavlik:
Oh wow, that's a handy github action now isn't it :) I guess probably yes, include the fix here? Sorry I hadn't thought of that detail when I said separate PR.

Biswa96:
I have added all the required fixes for mingw gcc and clang including the patch from previous pull request.

.github/workflows/mingw.yml
```diff
@@ -0,0 +1,32 @@
+name: mingw-build
+on: [push, pull_request]
```
rpavlik, line 2: This file needs a [REUSE](https://reuse.software) compliant copyright and license header, see the other files for the format.

src/tests/loader_test/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ add_executable(loader_test
 )
 openxr_add_filesystem_utils(loader_test)
 set_target_properties(loader_test PROPERTIES FOLDER ${TESTS_FOLDER})
-target_link_libraries(loader_test PRIVATE openxr_loader)
+target_link_libraries(loader_test PRIVATE openxr_loader XrApiLayer_core_validation)
```
rpavlik, line 26: why are you linking the validation layer to the loader test? Does this auto-copy stuff? I wouldn't expect anything to link at build time against an API layer, they get loaded at run time dynamically.

rpavlik (changes requested):
Looks pretty good, just minor suggestions!

Also if you could add a changelog file fragment in `changes/sdk/pr.296.gh.OpenXR-SDK-Source.md` - see the readme in that dir for info - that would really be helpful.

Biswa96:
It seems that the compiled DLLs have different names in msvc vs. mingw toolchain. That also affects the DLL name in corresponding import libraries. Changing any file name may break current use cases. This kind of breakage happened with Vulkan loader DLL before.

My sincere apology for this trouble. Feel free to use any of my changes anywhere you like 🙇

rpavlik:
Hmm, I don't think anyone is using the mingw build yet, it seems like it would be fine to fix the name there.

Biswa96:
mingw build is used here https://packages.msys2.org/package/mingw-w64-x86_64-openxr-sdk?repo=mingw64. The mingw dll name has 'lib' prefix but the msvc one does not. The patches can be found here https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-openxr

rpavlik:
aww crap, so they're carrying a patch to fix this themselves? In that case I don't really mind if the loader library has different names on mingw, not the end of the world. (We don't have any lib linking pragmas or anything, can name the library whatever you like, and it's not system installed on Windows. I suppose I should update the CMake find script, but the package produces a cmake package config file.) I do not know who is doing the mingw packaging or I'd reach out and try to get their stuff upstreamed. I'm glad someone's using it.

src/tests/loader_test/CMakeLists.txt
Biswa96: I get linker error like the following and though that would fix it.

```
ld.lld: error: undefined symbol: LogPlatformUtilsError(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&)
>>> referenced by src/tests/loader_test/CMakeFiles/loader_test.dir/__/__/common/filesystem_utils.cpp.obj:(utf8_to_wide(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&))
>>> referenced by src/tests/loader_test/CMakeFiles/loader_test.dir/__/__/common/filesystem_utils.cpp.obj:(utf8_to_wide(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&))
>>> referenced by src/tests/loader_test/CMakeFiles/loader_test.dir/__/__/common/filesystem_utils.cpp.obj:(wide_to_utf8(std::__1::basic_string<wchar_t, std::__1::char_traits<wchar_t>, std::__1::allocator<wchar_t>> const&))
```

rpavlik: I think the loader test is just missing an impl of LogPlatformUtilsError

Biswa96: Yes, after adding a dummy definition, the tests compiled successfully.

PR #397, opened 2023-04-19
Removed extraneous whitespace from xr.ml

satchelfrost:
Removed extraneous whitespace from xr.ml that was effecting  "PFN_xrCreateVulkanInstanceKHR", "PFN_xrCreateVulkanDeviceKHR", and "PFN_xrGetVulkanGraphicsDevice2KHR" in file openxr_platform.h.

### DIFF
--- a/changes/registry/pr.397.gh.OpenXR-SDK-Source.md
+++ b/changes/registry/pr.397.gh.OpenXR-SDK-Source.md
@@ -1,0 +1,1 @@
+registry: Remove extraneous whitespace from some commands.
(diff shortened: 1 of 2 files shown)


rpavlik:
This does make it match better, that's true. Was it causing a problem? I've considered adding that whitespace to all commands once I have the tool made to do so for structs.

satchelfrost:
It's not necessarily a problem, but whatever parses the xml to generate the code for openxr_platform.h takes into account that whitespace. For example the output on line 457 (I believe) of openxr_platorm.h is:

`typedef XrResult (XRAPI_PTR *PFN_xrCreateVulkanInstanceKHR)(XrInstance                           instance,  etc...`

Which carries over the whitespace. So either the python script that reads that xml should not take into account whitespace, or just don't have extra whitespace in the xml to begin with.

In the end, it's not hurting anything.

rpavlik:
ahh, sure. I've seen it do the same thing in the adoc. For some reason I thought the script auto-indented it when generating those.

PR #403, opened 2023-04-28
hello_xr: add debug output for Vulkan extensions from runtime vs. app

T-X:
Adding some extra debug output when "--verbose --graphics Vulkan(2)" is specified, to list which Vulkan instance and Vulkan device extensions were requested by the XR runtime and which were requested by the application.

I found this quite useful to have to get the current wineopenxr implementation running without Steam. Steam uses Windows registry keys at the moment to relay Vulkan extensions needed by the XR runtime to DXVK. Without Steam I need to populate the registry keys manually at the moment.


rpavlik (approved):
Thanks for this, and sorry for the delay! I made a few very minor changes, added a changelog fragment, and will merge this.

PR #404, opened 2023-05-09
Update AndroidManifest.xml

JaredMonkey:
Adding the necessary meta-data and intent for hello_xr to actually run on devices like Quest and Pico

### DIFF
--- a/src/tests/hello_xr/AndroidManifest.xml
+++ b/src/tests/hello_xr/AndroidManifest.xml
@@ -48,6 +48,14 @@
       android:theme="@android:style/Theme.Black.NoTitleBar.Fullscreen"
       android:exported="true"
       tools:ignore="NonResizeableActivity">
+      <!-- Required meta-data and inent category for Meta devices -->
+      <meta-data
+        android:name="com.oculus.supportedDevices"
+        android:value="quest|quest2|questpro" />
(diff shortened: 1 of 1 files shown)


rpavlik:
I'm a little undecided on this one: it seems like meta and pico should not be requiring these extra vendor specific metadata. I can understand requiring standard Android entries (like rumble permission) and the OpenXR specified intent filter, but these vendor specific ones feel like the opposite of what OpenXR should be doing. Will check with the wg in case you haven't already.

src/tests/hello_xr/AndroidManifest.xml
```diff
@@ -48,6 +48,14 @@
       android:theme="@android:style/Theme.Black.NoTitleBar.Fullscreen"
       android:exported="true"
       tools:ignore="NonResizeableActivity">
+      <!-- Required meta-data and inent category for Meta devices -->
+      <meta-data
```
utzcoz, line 52: Maybe it needs to moved into the `application` not `activity`? You can check the GitHub Action's building failure information.

utzcoz: You can check https://developer.oculus.com/documentation/native/android/mobile-native-manifest/ and https://developer-global.pico-interactive.com/document/unity/quickstart-configure-settings/ for Quest and PICO tutorial.

PR #413, opened 2023-07-05
OpenXR loader: add API layer discovery support.

quic-dengkail:
Loader currently did not support API layer disconvery yet. This patch supports this feature.

Loader will try to get all implicit/explicit API layers from broker supported by chosen active runtime and populate virtual API layer manifests from returned cursor. 
Here is authority and path used to find correct cursor: 
- Loader ---> broker

    - authority:org.khronos.openxr.runtime_broker
    - path:/openxr/major_ver/abi/[abi]/api_layer/[implicit:explicit]

Corresponding patch of broker part can be found here: https://gitlab.freedesktop.org/monado/utilities/openxr-android-broker/-/merge_requests/18

### DIFF
--- a/src/loader/android_utilities.cpp
+++ b/src/loader/android_utilities.cpp
@@ -185,30 +185,15 @@ static constexpr auto ABI = "x86_64";
 #error "Unknown ABI!"
 #endif
 
-/// Helper class to generate the jsoncpp object corresponding to a synthetic runtime manifest.
-class JsonManifestBuilder {
-   public:
-    JsonManifestBuilder(const std::string &libraryPathParent, const std::string &libraryPath);
-    JsonManifestBuilder &function(const std::string &functionName, const std::string &symbolName);
+static inline Json::Value makeMinimumVirtualRuntimeManifest(const std::string &libraryPath) {
+    Json::Value root_node(Json::objectValue);
 
-    Json::Value build() const { return root_node; }
-
-   private:
-    Json::Value root_node;
-};
-
-inline JsonManifestBuilder::JsonManifestBuilder(const std::string &libraryPathParent, const std::string &libraryPath)
-    : root_node(Json::objectValue) {
     root_node["file_format_version"] = "1.0.0";
     root_node["instance_extensions"] = Json::Value(Json::arrayValue);
     root_node["functions"] = Json::Value(Json::objectValue);
-    root_node[libraryPathParent] = Json::objectValue;
-    root_node[libraryPathParent]["library_path"] = libraryPath;
-}
-
-inline JsonManifestBuilder &JsonManifestBuilder::function(const std::string &functionName, const std::string &symbolName) {
-    root_node["functions"][functionName] = symbolName;
-    return *this;
+    root_node["runtime"] = Json::objectValue;
+    root_node["runtime"]["library_path"] = libraryPath;
+    return root_node;
 }
 
 static constexpr const char *getBrokerTypeName(bool systemBroker) { return systemBroker ? "system" : "installable"; }
@@ -259,9 +244,9 @@ static bool getCursor(wrap::android::content::Context const &context, jni::Array
     return true;
 }
 
-static int populateFunctions(wrap::android::content::Context const &context, bool systemBroker, const std::string &packageName,
-                             JsonManifestBuilder &builder) {
-    jni::Array<std::string> projection = makeArray({functions::Columns::FUNCTION_NAME, functions::Columns::SYMBOL_NAME});
+static int populateRuntimeFunctions(wrap::android::content::Context const &context, bool systemBroker,
+                                    const std::string &packageName, Json::Value &manifest) {
+    const jni::Array<std::string> projection = makeArray({functions::Columns::FUNCTION_NAME, functions::Columns::SYMBOL_NAME});
 
     auto uri = functions::makeRuntimeContentUri(systemBroker, XR_VERSION_MAJOR(XR_CURRENT_API_VERSION), packageName, ABI);
     ALOGI("populateFunctions: Querying URI: %s", uri.toString().c_str());
@@ -273,7 +258,7 @@ static int populateFunctions(wrap::android::content::Context const &context, boo
     auto functionIndex = cursor.getColumnIndex(functions::Columns::FUNCTION_NAME);
     auto symbolIndex = cursor.getColumnIndex(functions::Columns::SYMBOL_NAME);
     while (cursor.moveToNext()) {
-        builder.function(cursor.getString(functionIndex), cursor.getString(symbolIndex));
+        manifest["functions"][cursor.getString(functionIndex)] = cursor.getString(symbolIndex);
     }
 
     cursor.close();
@@ -324,15 +309,15 @@ int getActiveRuntimeVirtualManifest(wrap::android::content::Context const &conte
             // we found a runtime that we can dlopen, use it.
             dlclose(lib);
 
-            JsonManifestBuilder builder{"runtime", lib_path};
+            Json::Value manifest = makeMinimumVirtualRuntimeManifest(lib_path);
             if (hasFunctions) {
-                int result = populateFunctions(context, systemBroker, packageName, builder);
+                int result = populateRuntimeFunctions(context, systemBroker, packageName, manifest);
                 if (result != 0) {
                     ALOGW("Unable to populate functions from runtime: %s, checking for more records...", lib_path.c_str());
                     continue;
                 }
             }
-            virtualManifest = builder.build();
+            virtualManifest = manifest;
             cursor.close();
             return 0;
         }
--- a/src/loader/android_utilities.cpp
+++ b/src/loader/android_utilities.cpp
@@ -402,59 +402,67 @@ int getActiveRuntimeVirtualManifest(wrap::android::content::Context const &conte
     const jni::Array<std::string> projection =
         makeArray({active_runtime::Columns::PACKAGE_NAME, active_runtime::Columns::NATIVE_LIB_DIR,
                    active_runtime::Columns::SO_FILENAME, active_runtime::Columns::HAS_FUNCTIONS});
+    static bool hasQueryBroker = false;
+    static Json::Value runtimeManifest;
+    if (!hasQueryBroker) {
+        // First, try getting the installable broker's provider
+        systemBroker = false;
+        Cursor cursor;
+        if (!getActiveRuntimeCursor(context, projection, systemBroker, cursor)) {
+            // OK, try the system broker as a fallback.
+            systemBroker = true;
+            getActiveRuntimeCursor(context, projection, systemBroker, cursor);
+        }
+        hasQueryBroker = true;
 
-    // First, try getting the installable broker's provider
-    systemBroker = false;
-    Cursor cursor;
-    if (!getActiveRuntimeCursor(context, projection, systemBroker, cursor)) {
-        // OK, try the system broker as a fallback.
-        systemBroker = true;
-        getActiveRuntimeCursor(context, projection, systemBroker, cursor);
-    }
+        if (cursor.isNull()) {
+            // Couldn't find either broker
+            ALOGI("Could access neither the installable nor system runtime broker.");
+            return -1;
+        }
 
-    if (cursor.isNull()) {
-        // Couldn't find either broker
-        ALOGI("Could access neither the installable nor system runtime broker.");
+        cursor.moveToFirst();
+
+        do {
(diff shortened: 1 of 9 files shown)


quic-dengkail:
Hi @rpavlik, here is loader patch of API layer discovery, please help review, thanks!

src/loader/android_utilities.cpp
```diff
@@ -314,6 +365,134 @@ int getActiveRuntimeVirtualManifest(wrap::android::content::Context const &conte
...
     return 0;
 }
+
+static bool populateApiLayerManifest(std::string layerType, Cursor &cursor, std::vector<Json::Value> &layerRootNode) {
+    auto fileFormatVersion = cursor.getString(cursor.getColumnIndex(api_layer::Columns::FILE_FORMAT_VERSION));
+    auto name = cursor.getString(cursor.getColumnIndex(api_layer::Columns::NAME));
+    auto libDir = cursor.getString(cursor.getColumnIndex(active_runtime::Columns::NATIVE_LIB_DIR));
+    auto fileName = cursor.getString(cursor.getColumnIndex(api_layer::Columns::SO_FILENAME));
+    auto apiVersion = cursor.getString(cursor.getColumnIndex(api_layer::Columns::API_VERSION));
+    auto imllementationVersion = cursor.getString(cursor.getColumnIndex(api_layer::Columns::IMPLEMENTATION_VERSION));
```
rblenkinsopp: Typo:
```suggestion
    auto implementationVersion = cursor.getString(cursor.getColumnIndex(api_layer::Columns::IMPLEMENTATION_VERSION));
```

src/loader/android_utilities.cpp
```diff
@@ -314,6 +365,134 @@ int getActiveRuntimeVirtualManifest(wrap::android::content::Context const &conte
...
+    __android_log_print(ANDROID_LOG_INFO, TAG, "Got api layer: type: %s, name: %s, native lib dir: %s, fileName: %s, functions: %s",
+                        layerType.c_str(), name.c_str(), libDir.c_str(), fileName.c_str(), functions.c_str());
+
+    Json::Value rootNode(Json::objectValue);
+    rootNode["file_format_version"] = fileFormatVersion;
+    rootNode["api_layer"] = Json::objectValue;
+    rootNode["api_layer"]["name"] = name;
+    rootNode["api_layer"]["library_path"] = libDir + "/" + fileName;
+    rootNode["api_layer"]["api_version"] = apiVersion;
+    rootNode["api_layer"]["implementation_version"] = imllementationVersion;
```
rblenkinsopp: Typo:
```suggestion
    rootNode["api_layer"]["implementation_version"] = implementationVersion ;
```

src/loader/manifest_file.cpp
```diff
@@ -723,6 +723,112 @@ void ApiLayerManifestFile::AddManifestFilesAndroid(ManifestFileType type,
...
+    api_version.patch = 0;
+
+    if ((num_fields != 2) || (api_version.major == 0 && api_version.minor == 0) ||
+        api_version.major > XR_VERSION_MAJOR(XR_CURRENT_API_VERSION)) {
+        error_ss << "layer " << filename << " has invalid API Version.  Skipping layer.";
+        LoaderLogger::LogWarningMessage("", error_ss.str());
+        return;
+    }
+
+    uint32_t implementation_version = atoi(layer_root_node["implementation_version"].asString().c_str());
```
rblenkinsopp: We should prefer `strtol` to `atoi` there as there is a possibility of invalid input and `atoi`'s behaviour is undefined in this instance. `strtol` fails to `0` if the input is invalid:

```suggestion
    uint32_t implementation_version = atoi(layer_root_node["implementation_version"].asString().c_str());
    if (implementation_version == 0) {
        error_ss << "layer " << filename << " has invalid Implementation Version.  Skipping layer.";
        LoaderLogger::LogWarningMessage("", error_ss.str());
    }
```

src/loader/android_utilities.cpp
```diff
@@ -314,6 +365,134 @@ int getActiveRuntimeVirtualManifest(wrap::android::content::Context const &conte
...
+    } else {
+        __android_log_print(ANDROID_LOG_INFO, TAG, "functions field not existed!");
+    }
+
+    layerRootNode.push_back(rootNode);
+
+    return true;
+}
+
+static bool getApiLayerCursor(std::string layerType, wrap::android::content::Context const &context,
```
rblenkinsopp: I'm not sure this function name corresponds well to what it's doing. It's not returning a cursor to be used, it's using a cursor to populate a list of API layers, if present. I think something like `enumerateApiLayerManifests` or `populateApiLayerManifests` might be more appropriate?

The return value of this function also never appears to be used?

src/loader/runtime_interface.cpp
```diff
@@ -166,6 +166,25 @@ XrResult GetPlatformRuntimeVirtualManifest(Json::Value& out_manifest) {
...
+    if (context.isNull()) {
+        return XR_ERROR_INITIALIZATION_FAILED;
+    }
+    std::vector<Json::Value> virtualManifests;
+    if (0 != openxr_android::getApiLayerVirtualManifests(type, context, virtualManifests)) {
+        return XR_ERROR_INITIALIZATION_FAILED;
+    }
+    out_manifest = virtualManifests;
+    return XR_SUCCESS;
+}
```
rblenkinsopp: This should be wrapped with `XR_KHR_LOADER_INIT_SUPPORT` as it requires the LoaderInitData structure.
```suggestion
#if defined(XR_KHR_LOADER_INIT_SUPPORT)
XrResult GetPlatformApiLayerVirtualManifests(std::string type, std::vector<Json::Value>& out_manifest) {
    using wrap::android::content::Context;
    auto& initData = LoaderInitData::instance();
    if (!initData.initialized()) {
        return XR_ERROR_INITIALIZATION_FAILED;
    }
    auto context = Context(reinterpret_cast<jobject>(initData.getData().applicationContext));
    if (context.isNull()) {
        return XR_ERROR_INITIALIZATION_FAILED;
    }
    std::vector<Json::Value> virtualManifests;
    if (0 != openxr_android::getApiLayerVirtualManifests(type, context, virtualManifests)) {
        return XR_ERROR_INITIALIZATION_FAILED;
    }
    out_manifest = virtualManifests;
    return XR_SUCCESS;
}
#endif
```

src/loader/runtime_interface.hpp
```diff
@@ -31,6 +31,7 @@ class Value;
 //! Initialize loader, where required.
 XrResult InitializeLoader(const XrLoaderInitInfoBaseHeaderKHR* loaderInitInfo);
 XrResult GetPlatformRuntimeVirtualManifest(Json::Value& out_manifest);
+XrResult GetPlatformApiLayerVirtualManifests(std::string type, std::vector<Json::Value>& out_manifest);
```
rblenkinsopp:
```suggestion
#if defined(XR_KHR_LOADER_INIT_SUPPORT)
XrResult GetPlatformApiLayerVirtualManifests(std::string type, std::vector<Json::Value>& out_manifest);
#endif
```

rblenkinsopp (commented):
I've taken an initial pass over this with a few minor pieces, I want to double check the layer ordering as I think runtime shipped layers should be the lowest in the stack (just above the runtime) which I believe this is currently doing correctly but I want to double check.

rpavlik:
I did import this into OpenXR gitlab for internal discussion: https://gitlab.khronos.org/openxr/openxr/-/merge_requests/2888

If you can't access that you might need one of your colleagues to assist, since I know they don't let everyone there make a Khronos account.

quic-dengkail:
> I've taken an initial pass over this with a few minor pieces, I want to double check the layer ordering as I think runtime shipped layers should be the lowest in the stack (just above the runtime) which I believe this is currently doing correctly but I want to double check.

Yes, I can confirm this is exactly this patch will do. Runtime layers will be loaded above the chosen runtime, and thanks for the suggestions, I will modify patch accordingly.

src/loader/android_utilities.cpp
quic-dengkail: `enumerateApiLayerManifests` would be better, because `populateApiLayerManifests` has already been used to populate virtual manifest from returned cursor.

quic-dengkail:
Updates (9/27):

1. Functions/instance_extensions query pattern should be revised, better to use similar pattern as functions query in runtime discovery.(loader/broker/runtime side) ---> done

- Following paths will be used:
  - /openxr/[major_ver]/abi/[abi]/runtimes/active (existing)
  - /openxr/[major_ver]/abi/[abi]/runtimes/[package_name]/functions (existing)
  - /openxr/[major_ver]/abi/[abi]/api_layers/[implicit|explicit] (new in this MR - rename from api_layer )
  - /openxr/[major_ver]/abi/[abi]/api_layers/[package_name]/[layer_name]/functions (new in this MR)
  - /openxr/[major_ver]/abi/[abi]/api_layers/[package_name]/[layer_name]/instance_extensions (new in this MR)

- The proceedure for API layers would then be:
  -  Query active implict/explict layers by the /openxr/[major_ver]/abi/[abi]/api_layers/[implicit|explicit] path.

  -  For each layer get it's package_name and layer_name and check has_functions and has_instance_extensions columns.

  -  If has_functions is true, then query function mappings with /openxr/[major_ver]/abi/[abi]/api_layers/[package_name]/[layer_name]/functions

  - If has_instance_extensions is true, then query extensions and versions with /openxr/[major_ver]/abi/[abi]/api_layers/[package_name]/[layer_name]/instance_extensions
2. Adapt the protocol so that the existing authorities for the brokers are used, and so that both the system broker and installable broker can report API layers.(loader/broker/runtime side) ---> done
3. Remove ‘enableEnvironment’/‘disableEnvironment’ in API layers as they can have no effect on Android as EnvVars are not supported.(loader/broker/runtime side) ---> done
4. Multiple cases (case A, B, D, E, F1/F2/F3) in runtime/API layer discovery are not needed or have potential security issues to achieve ‘same-origin’ policy. (mainly loader side) ---> WIP

quic-dengkail:
Updates (10/18):

- Multiple cases (case A, B, D, E, G?, F1/F2/F3) in runtime/API layer discovery are not needed or have potential security issues to achieve ‘same-origin’ policy. (mainly loader side) --->DONE
  - Loader will firstly search runtime and decide where it comes from (JSON mainifest, system/installable broker), then it will search API layers accordingly.
    - Loader can get implicit/explicit API layers from apps or installable broker, when runtime is from installable broker. (exclude case A, B, F1)
    - Loader can get implicit/explicit API layers from apps or system broker, not installable broker, when runtime is from system broker. (include case C, exclude case G?, F2)
    - Loader can get implicit/explicit API layers from apps or system-installed json manifests, when runtime is from JSON mainifest(**active\_runtime\*\.json**). (exclude case D, E, G?, F3)

quic-dengkail:
12/7/2023: rebase code base to [OpenXR SDK 1.0.32.1 hotfix: Fix CMake build of loader on Universal Windows Platform](https://github.com/KhronosGroup/OpenXR-SDK-Source/commit/e93034ac7fd4b19c1c3b9b280824ec9cdba4562c)

quic-dengkail:
Rebase to latest: https://github.com/KhronosGroup/OpenXR-SDK-Source/commit/99256e9a637d733f713e6613e3f560be4ee22b2a (3/22)

quic-dengkail:
Rebase to latest: [9e9d023ffcc67037ee244f6f0bc3785370814c96](https://github.com/KhronosGroup/OpenXR-SDK-Source/commit/9e9d023ffcc67037ee244f6f0bc3785370814c96).  (4/18/2024)

- Rebase to loader release 1.1.36.

- Add "disable_environment" and "enable_environment" back on Android platform.

rpavlik:
OK! I was just importing this to GitLab, and I got a little carried away because I saw a few things to fix...

Anyway it is now a lot cleaner of a PR, less code duplication applied, etc, and there are individual commits for small atomic changes. Additionally, there are a handful of changes that are not dependent on the layer work and can be merged on their own.

I dropped the part of the change that cached the results of getting the API layers, as that seemed premature and probably unwise given that there are two authorities they could be pulled from. I did not drop the caching of the runtime chosen itself, but I did split it into a separate commit and I am inclined to drop it, as unrelated, more global state, etc.

This MR works fine (after my adjustments) with the current brokers out there (as in, does not grab any layers, but doesn't fall to pieces) but is a little needlessly loud because of an over-eager exception in the installable broker.

I did add a commit to require disable_sys_prop just like we require disable_environment, not sure if we want that or not. Also not sure if we want to validate the names of those props in any way, enforce a pattern or something?

PR #414, opened 2023-07-11
Hotfix for 1.0.28: Fix build of API layers on Windows on ARM: need bigobj

rpavlik:
1.0.28 release build failed because apparently we exceed the section limit but only on windows on arm. This just changes the build flags to fix that.


johnkearney:
Thanks Ryan. LGTM.

Issue #430, opened 2023-10-16
Can not use environment variable to control the loading of the layer on Android


RenfengLiu:
The functions to read the environment variable are not implemented for Android yet in https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/main/src/common/platform_utils.hpp#L352. But in the loader spec said https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/main/specification/loader/api_layer.adoc?plain=1#L475 layer can be enabled with the environment variable.

PR #431, opened 2023-10-16
Enable read environment variable on Android

RenfengLiu:
This fixs: #430

### DIFF
--- a/src/common/platform_utils.hpp
+++ b/src/common/platform_utils.hpp
@@ -327,14 +327,19 @@ static inline std::string PlatformUtilsGetSecureEnv(const char* name) {
 
 #include <sys/system_properties.h>
 
-static inline bool PlatformUtilsGetEnvSet(const char* /* name */) {
-    // Stub func
-    return false;
-}
+namespace detail {
 
-static inline std::string PlatformUtilsGetEnv(const char* /* name */) {
-    // Stub func
-    return {};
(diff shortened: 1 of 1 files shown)


rpavlik:
The underlying issue is that there's no way to set those environment variables from outside the app, so they don't actually serve the purpose intended. (though, there may not be implicit layer support on android at all yet? or maybe just from manifests?)

A port of the behavior is more likely to be using getprop with some prefix before the environment variable name, I suspect.

What's  your use case?

RenfengLiu:
There are two ways applications can set the environment variable:
- The wrap properties: https://source.android.com/docs/security/test/asan#using_the_wrap_property
- And the wrap script: https://developer.android.com/ndk/guides/wrap-script

RenfengLiu:
> The underlying issue is that there's no way to set those environment variables from outside the app, so they don't actually serve the purpose intended. (though, there may not be implicit layer support on android at all yet? or maybe just from manifests?)
> 
> A port of the behavior is more likely to be using getprop with some prefix before the environment variable name, I suspect.
> 
> What's your use case?

The use case is hoping that there's will same mechanism used in other platforms that can load the explicit layer or implicit layer when requested. The implict layers without the `enable_environment` works, but this loads the layer for all applications, which is not the desired behavior. 

Using `getprop` would be another option, but this might require the Android system reserve that prop.

PR #434, opened 2023-11-08
Add <uses-sdk> tag to android manifests for loader AAR.

rpavlik:
Without it, the manifest merger assumes something silly (<4) and adds permissions we don't want, which can cause some store rejections.


rpavlik:
Merged internally, in the latest release.

PR #439, opened 2023-12-01
Add missing header references in test Cmake

pfontain:
The test builds currently rely on the build loader dependency on the openxr headers for those headers to be pulled. This merge request makes the dependency explicit for tests, so that when the building the loader is off, the headers are still pulled.

### DIFF
--- a/src/tests/c_compile_test/CMakeLists.txt
+++ b/src/tests/c_compile_test/CMakeLists.txt
@@ -42,7 +42,7 @@ if(XR_USE_GRAPHICS_API_VULKAN)
     )
 endif()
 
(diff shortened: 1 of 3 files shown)


rpavlik:
How are you building anything else at all without the loader being built? It still needs the loader.

pfontain:
I use a pre-built loader here https://github.com/pfontain/OpenXR-SDK-Source/commit/2e7def6b72a98c293ff1b4d8d7e767060f60bdc3

rpavlik:
I think that, now that Meta has moved to the standard loader, this is no longer needed?

pfontain:
Probably. Thank you for the heads up.

PR #454, opened 2024-02-07
Option to skip building hello_xr test

lvonasek:
`hello_xr` test doesn't compile on Android out of the box. It requires defining "OS_LINUX" and including `glad` library.

Having `hello_xr` as a mandragory dependency complicates using this repository as a submodule and it is blocking for integrating OpenXR in Dolphin emulator.

### DIFF
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -15,8 +15,11 @@
 # limitations under the License.
 
 add_subdirectory(c_compile_test)
-add_subdirectory(hello_xr)
 add_subdirectory(list_json)
+
+if(NOT SKIP_HELLO_XR)
+    add_subdirectory(hello_xr)
+endif()
 
 if(NOT ANDROID)
     add_subdirectory(list)


lvonasek:
I do not understand what is wrong with the "Generate archive" flow. Could anyone help?

rpavlik:
The workflows here are messed up, I haven't had a chance to fix them. I assume it's probably not because of your change.

Is there a reason why you're not just using the artifacts from Maven Central that we publish and recommend?

(Also, you should just be able to disable tests for this as a submodule, and/or, use the OpenXR-SDK repo which is intended for submodule usage, and only contains the loader.)

Also it definitely should build for Android out of the box, though CMake won't make a full APK, the CMake is run by Gradle to build the native code. What is your build environment that you're having issues? (And where does it mention glad? I was pretty sure that change was reverted entirely before the last release due to an issue on Android and so shouldn't have been public.)

lvonasek:
> Is there a reason why you're not just using the artifacts from Maven Central that we publish and recommend?

There is a requirement to support both Android standalone and PCVR on Windows.

> (Also, you should just be able to disable tests for this as a submodule, and/or, use the OpenXR-SDK repo which is intended for submodule usage, and only contains the loader.)

I had to dig into the CitraVR integration which I use as a reference. OpenXR loader doesn't work on Meta Quest OS V62. They are using this workaround: https://github.com/amwatson/OpenXR-SDK-Source/commits/main/
(feel free to create an issue for this one)

---

However, thank you for pointing me on OpenXR-SDK repo. I will use it once it is compatible with the current Quest OS.

rpavlik:
yeah a fix equivalent to that will be in the next release.

PR #470, opened 2024-03-29
hello xr: change the ViewConfigurationType

jypeitao:
Change the ViewConfigurationType which app expected but runtime not supported to the one the runtime prefers the application to use.


rpavlik:
I think ideally we'd just have it as a command line option. In this case, the app is flexible enough to support just about any view config, but this is intended to be a two-step interaction in general: app gets the runtime preferences, and chooses the first one on the runtime's list that is also on the app's "known supported" list.

PR #475, opened 2024-04-25
Fix implicit API layers being loaded as explicit on Android

pembem22:
This replaces the hardcoded API layer type with an actual type parameter that's supplied to the function, otherwise implicit API layers are treated as explicit.


rpavlik:
Good catch, thank you!

rpavlik (approved):

PR #486, opened 2024-07-14
Fix CMake Warning FindPythonInterp and FindPythonLibs modules are removed

rickyjames35:
When running cmake configure on this project I get:
> CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
the cmake_policy command to set the policy and suppress this warning.

This is caused by usage of deprecated functionality `find_package(PythonInterp 3)` in the root `CMakeLists.txt`.

Relevant CMake documentation: https://cmake.org/cmake/help/latest/policy/CMP0148.html
> The FindPythonInterp and FindPythonLibs modules are removed.
These modules have been deprecated since CMake 3.12. CMake 3.27 and above prefer to not provide the modules.
 
Here's the new way of finding Python: https://cmake.org/cmake/help/latest/module/FindPython3.html#module:FindPython3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@
 cmake_minimum_required(VERSION 3.16)
 project(OPENXR)
 
-find_package(PythonInterp 3)
+find_package(Python3 COMPONENTS Interpreter)
 include(CTest)
 
 # Enable IDE GUI folders.  "Helper targets" that don't have interesting source code should set their FOLDER property to this
@@ -44,7 +44,7 @@ option(
     OFF
 )
 
-if(BUILD_FORCE_GENERATION AND NOT PYTHON_EXECUTABLE)
+if(BUILD_FORCE_GENERATION AND NOT Python3_EXECUTABLE)
     message(FATAL_ERROR "BUILD_FORCE_GENERATION requires Python")
 endif()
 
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -21,7 +21,7 @@
 cmake_minimum_required(VERSION 3.16)
 project(OPENXR)
 
-find_package(PythonInterp 3)
+find_package (Python3)
```
rickyjames35: I can also change this to `find_package (Python3 COMPONENTS Interpreter)` if people prefer for more clarity. But `find_package (Python3)` will default to this.
> If no COMPONENTS are specified, Interpreter is assumed.
https://cmake.org/cmake/help/latest/module/FindPython3.html#module:FindPython3

rickyjames35:
Seems that `src/CMakeLists.txt` has a couple places where the `PYTHONPATH` is being passed in. I was hoping this would be an easy fix but it might not be. @rpavlik do you have any thoughts?
https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/650f7514c16382828e363146b78330cc365cd83e/src/CMakeLists.txt#L426

maksim-petukhou:
When using Python3 cmake module, `Python3_EXECUTABLE` should be used instead of PYTHON_EXECUTABLE in `/CMakeLists.txt`, `/src/CMakeLists.txt` and `/include/openxr/CMakeLists.txt`. And there is no problem with `PYTHONPATH`.

rickyjames35:
Thanks @maksim-petukhov, your suggestion worked.

rpavlik:
Will this work as far back as CMake 3.16?

edit: ah, the deprecation started in 3.12, so apparently it is fine, oops. We were holding the line at 3.10 for a long time.

rpavlik (approved):
Thanks for this! I have squashed your commits, adjusted the commit message, added a changelog fragment, and fixed the CMake formatting (apparently the CMake format config file is not shipped in this repo?)

PR #490, opened 2024-07-23
Clarify xrEnumerateInstanceExtensionProperties for API layers

fredemmott:
Per https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#api-initialization , this function *can* be called before calling `xrCreateInstance`.

> This function may: be called before an instance has been created; implementations must: not assume an instance exists.

Concretely, a bug in past versions of the Ultraleap API layer crashed if an instance had not been created.


fredemmott:
Related docs PR: https://github.com/KhronosGroup/OpenXR-Docs/pull/173

rblenkinsopp:
I think this is a good clarification. I'd be happy for this to be merged.

PR #500, opened 2024-09-07
ci: Remove gradle-wrapper-validation job from GitHub Actions

utzcoz:
See
https://github.com/gradle/actions?tab=readme-ov-file#the-wrapper-validation-action.

From setup-gradle@v4, it performs wrapper validation for every execution, and the project doesn't need to run gradle-wrapper-validation separately.


rpavlik (approved):
Thank you!

PR #501, opened 2024-09-15
ci: Add GitHub Action for macOS building

utzcoz:
1. macos-14 for arm64 macOS.

Looks like humbletim/setup-vulkan-sdk's cache mechanism doesn't work for multiple jobs with the same OS but different architectures, like macOS 13 is x86_64 and macOS 14 is arm64. So this CL only adds macos-14 as runner OS and we can enable the macos-13 when we fixed the upstream cache supporting.

### DIFF
--- a/.github/workflows/macOS.yml
+++ b/.github/workflows/macOS.yml
@@ -1,0 +1,41 @@
+# Copyright (c) 2024, The Khronos Group Inc.
+# SPDX-License-Identifier: CC0-1.0
+
+name: macOS
+
+on:
+  push:
+    branches: [ main ]
+
+  pull_request:
+    branches: [ main ]
+
+permissions:
+  contents: read
+
+jobs:
+  macOS-build:
+    strategy:
+        matrix:
+          device: [
+            macos-14, # Tests Mac arm64
+          ]
+    runs-on: ${{ matrix.device }}
+
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/macOS.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: humbletim/setup-vulkan-sdk@v1.2.0
+        with:
+          vulkan-query-version: 1.3.290.0
+          vulkan-components: Vulkan-Headers, Vulkan-Loader
+          vulkan-use-cache: true
+      - name: Build
+        run: |
+          mkdir -p build/macos
+          cd build/macos
+          cmake -G "Xcode" ../..
```
rpavlik: looks like this maybe only generates, and doesn't build?

Also I typically use the `-S` and `-B` args to cmake these days to specify source and build dir. Avoids having to make directories, change directory, etc.

rpavlik (changes requested):
OK, this is awesome, thank you! I just have one little issue where I think it's not actually building, but otherwise this is great.

.github/workflows/macOS.yml
utzcoz: Thanks. I updated the command to build on macOS with xcodebuild.

utzcoz:
@rpavlik PTAL.

rpavlik (approved):
Looks good, thanks! I made a few minor tweaks.

PR #514, opened 2024-12-14
Add workflow to include pre-built loader on macOS

xwovr:
Here is an example run which triggered manually:
https://github.com/xwovr/OpenXR-SDK-Source/actions/runs/12325232718

Test steps:
1. Download the artifact `openxr_loader_macos.zip` from the above run
2. Unzip the artifact to `openxr_loader_macos` folder. And then `cd openxr_loader_macos`
3. Following the instruction in `README.md`, run `xattr -d com.apple.quarantine lib/libopenxr_loader.dylib`
4. Create new Unity project, import `com.unity.xr.openxr`, `com.unity.xr.interaction.toolkit` and `com.meta.xr.simulator` packages
5. Replace the `libopenxr_loader.dylib` in the project's "Packages/OpenXR Plugin/RuntimeLoaders/osx" folder with the one downloaded
6. Meta XR Simulator can be activated and launched successfully using the new OpenXR loader

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -1,0 +1,57 @@
+# Copyright 2021-2024, Collabora, Ltd.
+# SPDX-License-Identifier: CC0-1.0
+
+name: macOS builds
+on:
+  workflow_call:
+    inputs:
+      organizeAndRelease:
+        description: "Should we organize and release our artifacts?"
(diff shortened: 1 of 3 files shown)


rpavlik:
WG accepted this change to start publishing these artifacts, thanks for your work!

PR #517, opened 2024-12-20
Fix archive-sdk when running on wsl

danwillm:
When trying to run `maintainer-scripts/archive-sdk.sh` on wsl, it failed to build the headers with `No generator options for unknown target: openxr.h`. This seemed to be because there was an extra character at the end of the filename, so failed to find the key in `genOpts`.

This removes the whitespace at the end of the strings before passing it to `genxr.py` to get it working.

### DIFF
--- a/maintainer-scripts/archive-sdk.sh
+++ b/maintainer-scripts/archive-sdk.sh
@@ -70,6 +70,7 @@ makeSubset "$TARNAME" $(getSDKFilenames)
 
 # Read the list of headers we should generate, and generate them.
 while read -r header; do
+    header=$(echo "$header" | sed 's/[[:space:]]*$//')
     generate_spec include/openxr "$header" "$TARNAME"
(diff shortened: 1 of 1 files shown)


rpavlik:
Well that is an interesting failure mode - I assume you're running it on a clone on Windows so it used CRLF EOLs, but bash in WSL isn't stripping those of course.

Can we just do the trim in the python script instead? One less process invocation per file, etc.

danwillm:
Thanks - I did originally test just having the strip in python, but there's also places in the script that need the stripped name too, i.e. in add_to_tar when adding the files. Otherwise, it fails with `tar: include/openxr/openxr.h\r: Cannot stat: No such file or directory`.

Happy to make changes if you have any ideas, let me know.

rpavlik:
Ah, OK, that makes sense then.

PR #538, opened 2025-03-06
Destroy Pipeline and all ShaderModule

remmel:
- `vkDestroyShaderModule` : bug in the loop. Was always destory the 0th item
- `vkDestroyPipeline` : Pipeline::Release never call. Use destructor, like in the other structs

In order to avoid thoses warnings

```
VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0x4872eaa0 | vkDestroyDevice(): OBJ ERROR : For VkDevice 0x21a28d16d50[], Vk
ShaderModule 0xcfef35000000000a[] has not been destroyed.

validation layer: Validation Error: [ VUID-vkDestroyDevice-device-05137 ] Object 0: handle = 0xd10d270000000018, type =
VK_OBJECT_TYPE_PIPELINE; | MessageID = 0x4872eaa0 | vkDestroyDevice(): OBJ ERROR : For VkDevice 0x232cccfd950[], VkPipel
ine 0xd10d270000000018[] has not been destroyed.
```

### DIFF
--- a/src/tests/hello_xr/graphicsplugin_vulkan.cpp
+++ b/src/tests/hello_xr/graphicsplugin_vulkan.cpp
@@ -352,7 +352,7 @@ struct ShaderProgram {
         if (m_vkDevice != nullptr) {
             for (auto& si : shaderInfo) {
                 if (si.module != VK_NULL_HANDLE) {
-                    vkDestroyShaderModule(m_vkDevice, shaderInfo[0].module, nullptr);
+                    vkDestroyShaderModule(m_vkDevice, si.module, nullptr);
                 }
                 si.module = VK_NULL_HANDLE;
             }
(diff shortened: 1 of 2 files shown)


src/tests/hello_xr/graphicsplugin_vulkan.cpp
```diff
@@ -871,7 +871,7 @@ struct Pipeline {
         CHECK_VKCMD(vkCreateGraphicsPipelines(m_vkDevice, VK_NULL_HANDLE, 1, &pipeInfo, nullptr, &pipe));
     }
 
-    void Release() {
+    ~Pipeline() {
```
rpavlik: can you just call Release from the destructor instead of getting rid of it?

rpavlik (changes requested):
Just one minor tweak.

If you could also add a changelog fragment in changes/sdk/pr.538.gh.OpenXR-SDK-Source.md per the readme in there, that would be appreciated. If not I can try my hand at it.

rpavlik (approved):

PR #554, opened 2025-07-14
Include <iterator> directly in api_layer_interface.cpp

DKLoehr:
The newest version of LLVM's libc++ no longer includes <iterator> transitively, so we need to include it directly in order to have access to `std::back_inserter`.


rpavlik:
Thanks!

rpavlik (approved):

PR #556, opened 2025-07-28
Include <algorithm> directly in api_layer_interface.cpp.

matts1:
api_layer_interface fails to compile when using libcxx as a module.


rpavlik (approved):
ah, for std::copy and std::back_inserter. Yep, thanks!

PR #569, opened 2025-10-30
Export CMake configuration

Pandapip1:
This allows for dependents to find openxr-loader with `find_package(OpenXR)`

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -620,3 +620,16 @@ endif()
 if(BUILD_CONFORMANCE_TESTS)
     add_subdirectory(conformance)
 endif()
+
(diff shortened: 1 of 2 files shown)


rpavlik:
I could have sworn we already did that, doesn't the Windows binaries and the AAR both contain a full set of CMake config files?

Pandapip1:
It's possible it's a NixOS packaging issue but I couldn't find it in the output. I'll double-check.

EDIT: Yea, it appears that this is probably a packaging issue. Closing; sorry for the noise!

https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/ba4aec9686cb94c99a55f7ceba9768e9e35525c2/src/loader/CMakeLists.txt#L436

PR #577, opened 2025-12-23
Use the `QUITE` mode when looking for `Sanitizers`

corporateshark:
Prevent CMake warnings.


rpavlik:
OK, after some internal discussion, this looks like the cleanest way to handle this.

rpavlik:
Merged, thank you! (I merged locally so I could add a changelog fragment.)

PR #587, opened 2026-02-03
OpenXR 1.1.54 hello_xr sample fixes

emaschino:
A few fixes for the hello_xr sample that has been updated with OpenXR 1.1.54.


rpavlik (approved):
Thanks for this catch! Looks good. WG approved merge without further review.

PR #606, opened 2026-05-21
API dump: print XrPath values

xytovl:
The API dump shows numeric values of an XrPath, which are not convenient to track what is being called.

This change calls `xrPathToString` on each value and prints the result.
It appears that `PathToString` always returns XR_SUCCESS through the dispatch table, so I can't handle invalid paths correctly.

Sample new output:
```
XrResult xrSuggestInteractionProfileBindings
    XrInstance instance = 0x55d95352be10
    const XrInteractionProfileSuggestedBinding* suggestedBindings = 0x00007ffcda491fe0
    XrStructureType suggestedBindings->type = XR_TYPE_INTERACTION_PROFILE_SUGGESTED_BINDING
    const void * suggestedBindings->next = 0x0000000000000000
    XrPath suggestedBindings->interactionProfile = /interaction_profiles/bytedance/pico_neo3_controller
    uint32_t suggestedBindings->countSuggestedBindings = 0x44
    const XrActionSuggestedBinding* suggestedBindings->suggestedBindings = 0x55d9537aed50
    const XrActionSuggestedBinding* suggestedBindings->suggestedBindings[0] = 0x000055d9537aed50
    XrAction suggestedBindings->suggestedBindings[0].action = 0x55d95371a030
    XrPath suggestedBindings->suggestedBindings[0].binding = /user/hand/left/input/aim/pose
    const XrActionSuggestedBinding* suggestedBindings->suggestedBindings[1] = 0x000055d9537aed60
    XrAction suggestedBindings->suggestedBindings[1].action = 0x55d95371bd10
    XrPath suggestedBindings->suggestedBindings[1].binding = /user/hand/left/input/trigger/value
[...]
```

### DIFF
--- a/src/api_layers/api_dump.cpp
+++ b/src/api_layers/api_dump.cpp
@@ -233,7 +233,7 @@ XrInstance FindInstanceFromDispatchTable(XrGeneratedDispatchTable *dispatch_tabl
 }
 
 // Function to record all the API dump information
-bool ApiDumpLayerRecordContent(std::vector<std::tuple<std::string, std::string, std::string>> contents) {
+bool ApiDumpLayerRecordContent(std::vector<Argument> contents) {
     bool success = false;
     if (g_record_info.initialized) {
         std::unique_lock<std::mutex> mlock(g_record_mutex);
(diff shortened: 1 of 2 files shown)


rpavlik:
Hey this looks pretty handy. Any way you could print both the atom value and the string?

xytovl:
If we want to have both id and value, options are:
* include both in the "value" field, so we get something like "0: some name". This will be semantically incorrect, the html output will have a single `<div class="val">0: some name</div>`. 
* do as if it were a struct with id and value fields, html output will be like
```html
<details class="data">
  <summary>...</summary>
  <div class="data">
    <div class="type">int</div>
    <div class="var">id</div>
    <div class="val">0</div>
  </div>
  <div class="data">
    <div class="type">string</div>
    <div class="var">value</div>
    <div class="val">some value</div>
  </div>
</details>
```
* extend the return type with an optional "decoded" value so that the generated html could separate the identifier and the logical value

rpavlik:
ah, I wasn't thinking about folks using the HTML as a structured output. Probably the latter option is the best solution.

PR #612, opened 2026-07-19
Fix Vulkan resource cleanup in hello_xr

### LINKED ISSUE #211: hello_xr does not gracefully clean up Vulkan device/instance
`VulkanGraphicsPlugin` should destroy the `VkDevice` and `VkInstance` it owns on destruction rather than leaking them. This allows the Vulkan validation layers to detect resource leaks, which is useful for finding bugs both in hello_xr itself and in runtimes that use hello_xr with the validation layers as a smoke test for correctness.

HussainAther:
## Summary

Add explicit cleanup for Vulkan resources owned by `VulkanGraphicsPlugin`.

The destructor now:

* waits for the Vulkan device to become idle
* releases swapchain image and rendering resources
* destroys the draw semaphore
* resets the memory allocator
* destroys the Vulkan device
* destroys the debug messenger
* destroys the Vulkan instance
* clears the associated Vulkan handles

This prevents `hello_xr` from leaking its Vulkan device and instance and allows validation layers to report resource leaks more accurately.

Fixes #211.

## Testing

* `cmake --build build`
* `ctest --test-dir build`

Result:

```text
100% tests passed, 0 tests failed out of 1
```

### DIFF
--- a/src/tests/hello_xr/graphicsplugin_vulkan.cpp
+++ b/src/tests/hello_xr/graphicsplugin_vulkan.cpp
@@ -468,6 +468,60 @@ class VulkanSwapchainImageData : public SwapchainImageDataBase<XrSwapchainImageV
 struct VulkanGraphicsPlugin : public IGraphicsPlugin {
     VulkanGraphicsPlugin() { m_graphicsBinding.type = GetGraphicsBindingType(); };
 
+    ~VulkanGraphicsPlugin() override {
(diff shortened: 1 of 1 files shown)


johnkearney:
This looks like a good change to me but as I compare it to the equivalent code in the conformance test suite I'm seeing divergence. Generally we've tried to keep these inline - if there's a way to keep them more similar that'd be great. But generally LGTM.

PR #616, opened 2026-09-15
hello_xr: Fix incorrect printf format

mbucchia-valve:
The bug can easily be reproduced when building for 32-bit targets (either Windows or Linux). eg:

```sh
$ mkdir out32
$ cd out32
$ CFLAGS="-m32" CXXFLAGS="-m32" cmake ..
$ make -j16
$ src/tests/hello_xr/hello_xr -g vulkan
[15:26:36.003][Info   ] Press any key to shutdown...
[15:26:36.014][Info   ] Available Layers: (0)
[15:26:36.014][Info   ] Depth submission supported (XR_KHR_composition_layer_depth)
[15:26:36.025][Info   ] Instance RuntimeName=SteamVR/OpenXR RuntimeVersion=2.18.0
Segmentation fault (core dumped)
```

With patch:

```
$ src/tests/hello_xr/hello_xr -g vulkan -v
[15:44:23.504][Info   ] Press any key to shutdown...
[15:44:23.514][Verbose] Available Extensions: (40)
[...]
[15:44:23.525][Verbose] Using system 1168881766287540733 for form factor XR_FORM_FACTOR_HEAD_MOUNTED_DISPLAY
[...no crash...]
```


rpavlik:
Looks good, can you add a changelog fragment in changes/sdk? Something starting like "hello_xr: Fix: format string size mismatch"

rpavlik (changes requested):
just looking for a changelog frag

mbucchia-valve:
Will do. I also just realized that the prefix I meant to use was "hello_xr:" not "openxr:" so I'll fix that too.

rpavlik (approved):

